PR #14, opened 2024-08-23
Add support for pyproject.toml standard

### LINKED ISSUE #13: Add pyproject.toml support

3x10RaiseTo8:
Fixes #13


JoshuaC215:
Thanks @3x10RaiseTo8 for working on improving the package management!

I believe the best practice for python now is to use a tool besides poetry, since it doesn't conform to the pyproject.toml standard

https://www.stuartellis.name/articles/python-modern-practices/#use-a-project-tool
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/

I have been using `uv` lately (https://github.com/astral-sh/uv) since it is so dang fast

Would you be willing to convert this to use a standard pyproject.toml ? And in that case, we could also remove `requirements.txt` and `test-requirements.txt` in the same PR - it would require an update in the Dockerfiles and test.yml workflow file as well.

If this is more than you want to take on, I can take a pass at it before too long.

3x10RaiseTo8:
Outstanding reads! Migrating...

Issue #30, opened 2024-09-05
Token-level streaming with Gemini


peterkeppert:
I was testing the toolkit and discovered that while token-level streaming was working great with OpenAI models, it wasn't working properly with Gemini (running on VertexAI) and I was getting only the final message from the model. For some reason, _on_llm_new_token()_ in _TokenQueueStreamingHandler_ was never called and I didn't find a reason why. I doubt the issue was with my setup, as I just switched the models in _research_assistant.py_ and double-checked _streaming=True_ is present. I am using Langchain 0.2.15.

I was able to fix the issue and have token-level stremaing working consistently for both models by switching from _.astream()_ to _.astream_events()_ ([docs](https://langchain-ai.github.io/langgraph/concepts/low_level/#astream_events-for-streaming-tokens-of-llm-calls)) in _message_generator()_ and catching _on_chain_end_/_ChannelWrite_ events for messages updating the state and _on_chat_model_stream_ events for token stream. However this is quite a conceptual change for a relatively small issue (not to mention that event streaming API is still in Beta).

Did anybody have the same issue with Gemini and perhaps found a better way to fix it?

saltwicks:
I'm experiencing the same issue with a custom langchain ChatModel. Would you be able to share the fix you implemented by changing over to astream_events and modifying the callback?

JoshuaC215:
Yeah feel free to post a PR even in draft, might be a useful reference for others. I won’t be able to take a close look before next week. I haven’t gotten the current approach working other than with OpenAI. I’d be curious how it works with other OpenAI-compatible APIs (like FastChat) but not at all opposed to a different approach if it provides better coverage of model providers. Thank you!

peterkeppert:
I have attached a draft PR. For the moment I didn't find a better way to select relevant _on_chain_end_ events. I guess it might get more straightforward as event streaming API evolves. Also, it is tested only on a simple graph from within the toolkit, I will have an opportunity to check it on a more complex example hopefully next week.

JoshuaC215:
BTW, it's probably a separate PR but if you have simple code for running this with a Gemini model would love to include that in the base repo as well!

JoshuaC215:
Resolved with #32 thank you again for opening this and getting the fix started!

If you have a simple way to add Gemini as an example and want to create another PR please do

PR #32, opened 2024-09-06
Switch to astream_events to support token-level streaming with Gemini

peterkeppert:
Draft PR for #30

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -9,13 +9,13 @@ services:
       - .env
     develop:
       watch:
-        - path: agent/
+        - path: src/agent/
           action: sync+restart
           target: /app/agent/
-        - path: schema/
+        - path: src/schema/
           action: sync+restart
           target: /app/schema/
-        - path: service/
+        - path: src/service/
           action: sync+restart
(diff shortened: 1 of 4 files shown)


JoshuaC215:
I think the CI issue is due to lint running on python 3.9 and anext() being added in 3.10. I can fix that, it seems reasonable to set that minimum version.

Will take a closer look at this in the next few days.

JoshuaC215:
Made some progress on cleaning this up but there is still some weirdness with filtering the `on_chain_end` events - seems like the langchain implementation changes across versions, and makes it harder to not leak service details into the agent. I need to take a closer look and think about it some more.

I think with this change we can also remove the asyncio.Queue usage which will simplify things a bit too, which is nice.

I still want to incorporate this but might be later this week or next week before I get a chance to dig in. Also need to do the langchain v0.3 release update (which should solve some other problems too :tada: )

JoshuaC215 (approved):
Looking good now 💪

JoshuaC215:
Thank you again @peterkeppert for this awesome contribution!

PR #37, opened 2024-09-23
Add Gemini to examples

peterkeppert:
Hello @JoshuaC215 , as per discussion in #30 , I have created a pull request to add Gemini as an example. For this to work, you 

GOOGLE_API_KEY environment variable is required if model is to work properly, you can get Gemini API key in Google AI Studio, [see here](https://ai.google.dev/gemini-api/docs/api-key).

PR #41, opened 2024-09-25
Add claude model example

peterkeppert:
I have created a pull request to support Claude (Anthropic) models. Without streaming, it was as straightforward as #37 . With streaming, it was slightly more complicated. Until now, all example models' streamed and final messages had a string content. 

However, Langchain's BaseMessage allows `content: Union[str, List[Union[str, Dict]]]` and our `schema.ChatMessage` content allows only `str` and Anthropic uses the `Dict` variant in streaming mode. @JoshuaC215 I was already wondering about this before,  @JoshuaC215 , could you please shed some light on the reasons for this decision?

For now, I have created a helper method in schema to convert the content to string if necessary (applies both to streamed tokens and final messages). I would note that:
- It can parse only textual content for now. For any other output types (images, tool calls data in content - as happens for instance with Claude models), it will produce empty content. The rest of message values are not altered.
- I am not sure this is the best solution. It seems reasonable to allow ChatMessage content the same spectrum of types as Langchain BaseMessage content has and perform this conversion exclusively on GUI.

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -8,12 +8,13 @@ from fastapi import FastAPI, HTTPException, Request, Response
 from fastapi.responses import StreamingResponse
 from langchain_core._api import LangChainBetaWarning
 from langchain_core.runnables import RunnableConfig
+from langchain_core.messages import AIMessageChunk
 from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
 from langgraph.graph.graph import CompiledGraph
 from langsmith import Client as LangsmithClient
 
 from agent import research_assistant
-from schema import ChatMessage, Feedback, UserInput, StreamInput
+from schema import ChatMessage, Feedback, UserInput, StreamInput, convert_message_content_to_string
 
 warnings.filterwarnings("ignore", category=LangChainBetaWarning)
 
@@ -108,7 +109,7 @@ async def message_generator(user_input: StreamInput) -> AsyncGenerator[str, None
                 # LangGraph re-sends the input message, which feels weird, so drop it
             if chat_message.type == "human" and chat_message.content == user_input.message:
                 continue
-            yield f"data: {json.dumps({'type': 'message', 'content': chat_message.dict()})}\n\n"
+            yield f"data: {json.dumps({'type': 'message', 'content': chat_message.model_dump()})}\n\n"
 
         # Yield tokens streamed from LLMs.
         if (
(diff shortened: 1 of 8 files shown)


src/service/service.py
```diff
@@ -121,7 +122,10 @@ async def message_generator(user_input: StreamInput) -> AsyncGenerator[str, None
                 # Empty content in the context of OpenAI or Anthropic usually means
                 # that the model is asking for a tool to be invoked.
                 # So we only print non-empty content.
-                yield f"data: {json.dumps({'type': 'token', 'content': content})}\n\n"
+                if isinstance(content, str):
+                    yield f"data: {json.dumps({'type': 'token', 'content': content})}\n\n"
+                elif isinstance(content, AIMessageChunk):
+                    yield f"data: {json.dumps({'type': 'token', 'content': convert_message_content_to_string(content.content)})}\n\n"
```
JoshuaC215, line 128: I think there's actually a bug here, because `content` is the List of dict, not an `AIMessageChunk`, so this always returns false.

I fixed it by getting rid of the if/else and replacing this section with one line:

```python
yield f"data: {json.dumps({'type': 'token', 'content': convert_message_content_to_string(content)})}\n\n"
```

It seems to work as expected after I make that change.

JoshuaC215:
Thanks for this. Yeah on the ChatMessage question - was trying to keep it simple with just the `str`. But maybe it backfired on me.

I introduced the whole abstraction to minimize the crossover for langchain in the client, and make it easier if someone wants to adapt this to another framework like LlamaIndex. But the dependency is still there, until / unless someone adds support for manually deserializing in the client.

Rather than doing that, I wonder if it's better to just remove this ChatMessage abstraction and just use LangChain base messages directly in the interface. Guessing it would remove / simplify a lot of code and some complexity that isn't adding value (and we're layering in more complexity with this fix so I agree it smells funny)

What do you think?

src/service/service.py
peterkeppert: You are correct, thanks for catching this!

src/service/service.py
```diff
@@ -121,7 +121,7 @@ async def message_generator(user_input: StreamInput) -> AsyncGenerator[str, None
                 # Empty content in the context of OpenAI or Anthropic usually means
```
peterkeppert, line 121: When retesting the fix, I noticed another issue, which I will have to look at more closely. As opposed to OpenAI models, when Claude uses tool calls, `content` is not empty, but has a value like this: `[{'id': 'toolu_01H6DqC7db6kEDMthUa7LrPx', 'input': {}, 'name': 'WebSearch', 'type': 'tool_use', 'index': 0}]` . The empty content check doesn't seem sufficient.

peterkeppert:
Thanks for the clarification. I agree with the reasoning and would definitely be in favor of removing the ChatMessage abstraction. I think besides simplifying the codebase, it makes it easier for people who are already accustomed to LangChain message types to adapt the toolkit, as they won't have to start using an additional custom message format.

Would you prefer to do it in a separate pull request? It's quite a big change impacting tests as well.

JoshuaC215:
Yeah I think removing ChatMessage is 1000% a separate pull request haha

src/service/service.py
JoshuaC215: Oh interesting, I hadn’t noticed that. Do you think we need to fix here before merging? Have a fix in mind?

peterkeppert: I was thinking which approach will be best, if to keep only text content or to only filter tool use content and keep the rest. I settled on the latter. When ChatMessage abstraction is removed and `convert_message_content_to_string()` goes with it, UI will have to adapt to the possibility of streamed content not being just strings, but also dictionary content items (possibly even images).

JoshuaC215 (approved):

Issue #38, opened 2024-09-23
Background task UI notification


peterkeppert:
I have been using the toolkit as a template to build a RAG-based agent. This agent performs 2-3 background tasks before generating a visible response for the user. Background tasks either retrieve structured output from LLM that informs agent's decision making or perform programatic tasks, like document retrieval from vector DB. Each of the steps is fairly fast, but together, before the streamed tokens from the final response begin to appear, the waiting time is easily >30s. It might not be that much, but I can easily imagine a longer-running background task (like waiting for human input or performing complicated retrieval).

I looked around and couldn't find a good solution how to notify the user on the progress of these tasks and make the app more interactive. Then I was thinking to repurpose tool calling functionality, but in the end decided against, because tool calls work differently - they basically start and end in LLM + are tied to given AI response message, whereas for the background tasks, LLM is just an optional part + they usually precede response generation step.

In the end, I made it work by creating a new TaskMessage based on Langchain's ChatMessage, extended Agent Service Toolkit's ChatMessage to support task-oriented fields and used custom events to stream the notifications. With small additions to UI, background tasks now appear in dedicated container in between the messages, where they in my opinion logically belong (see screenshot).

![background_tasks](https://github.com/user-attachments/assets/47193e97-8a92-428b-8b2f-ea574268d4ad)

Is this something somebody had to tackle as well and perhaps has a better solution at hand? @JoshuaC215 is this a functionality that you would wish to incorporate into the toolkit or would it bloat the template unnecessarily?

FDA-1:
I've recently started using this template, and this addition is exactly what I've been thinking of building. IMO, it would make perfect sense to add to the template.

JoshuaC215:
Nice! This sounds cool

I wonder if using something like `langchain_core.callbacks.adispatch_custom_event` would be the right approach for these progress updates, similar to [here](https://langchain-ai.github.io/langgraph/how-tos/streaming-events-from-within-tools-without-langchain/).

The `st.status` container used to render the tool output also supports [updating the label](https://docs.streamlit.io/develop/api-reference/status/st.status#statuscontainerupdate) with intermediate calls. So if you have multiple serial steps, another approach instead of showing multiple containers would be to keep one "running" container and update the label to show the current step

I am nervous about bloat but open to this if it doesn't require a huge refactor or a ton of additional code or branching logic. If we find it's too much, also happy to link to a separate branch / PR from the main README or something so it's more discoverable. What do you think?

[...]

PR #42, opened 2024-09-25
Background task notifications

peterkeppert:
PR for #38 I chose the safety check as example implementation.

I have a few points I would like improve here, but let's first see if the general idea is sound:
- Adding task-specific fields to ChatMessage bloated the class a lot. Considering there are already fields specific to Human, AI and Tools messages, maybe it would be best to refactor so that there are specific ChatMessages mirroring Langchain messages.
- As for changes in `research_agent.py`, I am not very happy that we would have to remember to remove task messages basically every time we want to use state["messages"] as model input. However, we need a place where all the messages including the Task messages are listed in order they were created, so that we can recreate the chat history including background task notifications.
- I also don’t like the fact that TaskMessage is based on Langchain’s ChatMessage. BaseMessage would be more logical choice of base class, but then TaskMessage wouldn’t satisfy AnyMessage checks. I think that Langchain could support some kind of task message even itself. Do you think it makes sense to propose creation of such message type?

@JoshuaC215 please review the changes and let me know your thoughts.


JoshuaC215:
Hmmm so here are my quick and candid thoughts on this:
- I like the Task abstraction
- I like the nice isolated handling in service.py
- I don't like the extension of ChatMessage to support TaskMessage and all the extra fields / code around that. I wonder if removing ChatMessage and just using LangChain types as mentioned in the other PR would help? Based on my experience with LangChain I doubt they will add this as a type, certainly not quickly without a strong use case.
- I really don't like having to remove the tasks in every use of `state["messages"]`, as you mentioned. Is there a way to store it under a different key in state? 🤔 
- I don't like the modifications in research_assistant.py and making that more complex to show this use case, since I think it's a bit niche. I wonder if it makes sense to add a simple version of the multiple agent support (kinda discussed in https://github.com/JoshuaC215/agent-service-toolkit/discussions/28) and then include this as a separate agent?

It might also be that this is just too complex to include as a niche case in the main repo (unless we see more discussion / demand for it) and we shouldn't work too much harder to try to force it in. I hope this doesn't land too discouraging and sorry if so!

antonioalegria:
Just jumping in on the conversation of removing the ChatMessage (also mentioned in #41), I do think ChatMessage is a useful abstraction. It simplifies and encapsulates what the ChatMessage should be in the presentation layer. It makes that layer much cleaner, even if it makes the service a bit more convoluted.

I think Langchain does not have a strong enough message abstraction as they can vary in structure inside depending on the model (as we see with the Anthropic models) and other context. Just my 2 cents.

peterkeppert:
@antonioalegria I agree that Langchain's Message content field isn't the easiest to work with. The typing of ` Union[str, List[Union[str, Dict]]]` introduces quite an overhead every time one wishes to parse it (or perhaps I am unaware of preexisting tooling). I wonder whether Langchain itself won't in the end standardize it on the List[Dict] format and simply apply `type = "text"` on items that would otherwise be strings. We could do it ourselves in the ChatMessage class. This would definitely be better than the current `str` content that prevents us from sending for example images to the UI altogether.

However, personally, I still favor removing ChatMessage class. It seems to me more in line with the purpose of the repository to provide a starting point for anybody willing to implement their own AI agent with Langgraph. By removing this class, we are giving up on trying to solve the issue of Langchain's imperfect message abstraction layer and simply showing how one can use it to implement an end-to-end agent with some limited UI capabilities (surely we won't demonstrate handling of all possible variations of message `content`). It removes some complexity from the code instead allowing us to focus on agent-related functionality, which I would be expecting this kind of repository to solve and leaving this particular issue to perhaps another project with heavier focus on UI.

That said, I believe the final decision rests with maintainer. @JoshuaC215 , please let us know how you decide. If you decide to remove the ChatMessage class, I think it would be best if you could open an issue for it, summarizing the reasons for the change. I think I would have some time next week to create a draft PR. Once we have this sorted out, I would finalize the background task notifications pull request.

JoshuaC215:
I wrote up the issue as requested, let me know what y'all think.

peterkeppert:
Since we solved the chat message abstraction question, I resolved conflicts and added the "detailed_messages" state key. I have also added the simple version of the multiagent support as discussed in #28 , since I had it already developed. If you want to switch frontend to different backend agent, you need to switch URLs in client.py. Please review :)

What remains is to write up Lanchain issue for new message type, I will reference this thread.

src/client/client.py
```diff
@@ -50,7 +50,7 @@ async def ainvoke(
             request.model = model
         async with httpx.AsyncClient() as client:
             response = await client.post(
-                f"{self.base_url}/invoke",
+                f"{self.base_url}/invoke/research-assistant",
```
antonioalegria, line 53: Client should not be tied to the research-assistant, especially not in so many places, otherwise it will be harder to use for other agents we might implement.

The agent may be a path parameter. Also would make more sense to make this part of a separate changeset.

src/schema/schema.py
```diff
@@ -98,6 +104,25 @@ class ChatMessage(BaseModel):
         description="Original LangChain message in serialized form.",
         default={},
     )
+    task_name: str | None = Field(
```
antonioalegria, line 107: to clean up ChatMessage, we could add a Task class with these fields and ChatMessage could have a task optional field.

src/schema/schema.py
```diff
@@ -128,6 +153,17 @@ def from_langchain(cls, message: BaseMessage) -> "ChatMessage":
                     original=original,
                 )
                 return tool_message
+            case TaskMessage():
+                task_message = cls(
+                    type="task",
```
antonioalegria, line 158: Make TaskMessage have type=Task in the class definition, that way you don't need to pass the type here

src/schema/schema.py
```diff
@@ -171,6 +207,93 @@ class Feedback(BaseModel):
...
+class TaskMessage(LangchainChatMessage):
+    """Message from a background task.
+
+    TaskMessages are returned by system to inform user about progress and results
+    of background tasks.
+    """
+
+    role: str = "system_task"
+    """The role of the speaker."""
+
```
antonioalegria, line 219: set type to Literal["task"] = "task"

src/service/service.py
```diff
@@ -98,29 +99,50 @@ async def invoke(user_input: UserInput) -> ChatMessage:
         raise HTTPException(status_code=500, detail=str(e))
 
 
-async def message_generator(user_input: StreamInput) -> AsyncGenerator[str, None]:
+@app.post("/invoke/research-assistant")
```
antonioalegria, line 102: research-assistant should not be in spread across the service code, only in the setup code (lifespan).

antonioalegria:
> Since we solved the chat message abstraction question, I resolved conflicts and added the "detailed_messages" state key. I have also added the simple version of the multiagent support as discussed in #28 , since I had it already developed. If you want to switch frontend to different backend agent, you need to switch URLs in client.py. Please review :)
> 
> What remains is to write up Lanchain issue for new message type, I will reference this thread.

I do think the multiagent code should be in a separate changeset, and the service and client code should be as unaware as possible as to what agent is setup. In the `main` branch today that is totally transparent in the client and in the server it is contained in the lifespan setup, as it should be.

My suggestion is to move the multi-agent support to another issue/pull-request so that approaches could be discussed.

src/schema/schema.py
peterkeppert: I don't think this will work for two reasons:

1. ChatMessage `type` is not the same as Langchain BaseMessage `type`. They are distinct fields. AIMessage, HumanMessage and ToolMessage have their `type` set in the respective Langchain classes, but we still have to set ChatMessage's type manually as in these other cases. We could write everywhere:
```cls(
  type=message.type,
  ...
```
However, I am not sure of the value of this change.

2. TaskMessage is based on Langchain's `ChatMessage` which has `type` already set to `chat`. I couldn't base it on Langchain's `BaseMessage`, where I could set `type` arbitrarily, because `TaskMessage` needs the message to pass `AnyMessage` Langchain type checks, which Langchain and Langgraph rely on when working with messages. Using `BaseMessage` as base class would fail `AnyMessage` checks.

src/service/service.py
peterkeppert: Done, removed multiagent implementation

src/client/client.py
peterkeppert: Done, removed multiagent implementation

peterkeppert:
Thanks for suggestions, multiagent implementation has been removed altogether, let's wait for progress on #28 . Instead, I added instructions to readme file.

src/schema/schema.py
peterkeppert: This looks appealing, but I think it breaks the proposed abstraction.

`Task` class models a background task. As the task progresses (when it starts, finishes, or when it produces data during its execution), it emits TaskMessages, which (similar to how HumanMessages or AIMessages are handled) are then converted to ChatMessages, which serve as universal communication tool between server and client.

That said, I believe placing `Task` itself into `ChatMessage` is wrong, as this way `ChatMessage` would contain a reference to background Task, which can progress during ChatMessage's lifetime. This contradicts ChatMessage's purpose to communicate a single notification produced by Task execution at given point in time.

JoshuaC215:
I left some thoughts on how to move this PR forward [here](https://github.com/JoshuaC215/agent-service-toolkit/issues/49#issuecomment-2439652701)

I'm hoping I'll have some time soon (in the next week or so, maybe sooner) to work on some of the changes that will unblock this.

JoshuaC215:
Spent some more time looking at this. Here's a commit of how I'd want to roughly approach merging this after #75 is merged. Open to discussion.

https://github.com/JoshuaC215/agent-service-toolkit/commit/c5c7a2ca5fe4d823032be7c1c0c3761d3d0bc790#diff-b678273f64e8e9df8f3a41b23f9fba9ca9e4809b93b53843fb1c370a86e69110

^ I tested this version and it seems to work! :tada:

I think we could do the changes in service.py in a generic way and maybe even put this in the main streamlit app (although I'm a bit hesitant about that 😅 for complexity reasons, but maybe its OK).

One thing to call out, it wasn't obvious to me why we'd need the `"detailed_messages"` thing so I removed that. But open to discuss. It gets lost in history but I think that's OK? I don't really want to add this detailed_messages thing everywhere. Alternatively, maybe going back to just storing them in messages and filtering out of the LLM calls is the way to go.

JoshuaC215:
Sounds good - I guess when I was trying it I wasn't seeing the use case for `"detailed_messages"` but maybe it makes sense in a more complex app. Anyway sounds like this will work for now. Happy to discuss that more after it's merged if you think there's a case to include it in the main toolkit.

Will revisit after #75 is merged in the next few days (hopefully). I want to get some review from other contributors since it's a big change. If you have any feedback on that PR (or other changes I made yesterday) please feel free to share!

src/streamlit_app.py
```diff
@@ -262,6 +262,55 @@ async def draw_messages(
...
+                            ":green[completed successfully]"
+                            if task_data.result == "success"
+                            else ":red[ended with error]"
+                        )
+                        status.write(f"Task **{task_data.name}** {result_text}. Output:")
+                status.write(task_data.data)
+                status.write("---")
+                task_messages[task_data.run_id] = msg
+                state = "complete"
+                for message in task_messages.values():
```
JoshuaC215, line 302: `message` isn't actually referenced here. I was looking through this whole section of code and I'm not sure it works the way you intended - probably can use more cleanup

JoshuaC215:
I merged all the pre-requisite changes including the general implementation for CustomData

Here's my pass on adding a bg task agent:
https://github.com/JoshuaC215/agent-service-toolkit/compare/main...bg-task-agent

I didn't clean up the code in the streamlit app marked above ^

Feel free to adapt this PR / open a new one. Or we can just use my branch if you prefer.

JoshuaC215:
Moved to #81

PR #43, opened 2024-09-30
Also allow feedback of 1 star

dietmar:
Currently, clicking 1 star for feedback results in `st.feedback()` returning `0` (see https://docs.streamlit.io/develop/api-reference/widgets/st.feedback), and this `if` is not entered:
https://github.com/JoshuaC215/agent-service-toolkit/blob/fbf072de05a8a5ea911fc484a0e8c0516dab8ea4/src/streamlit_app.py#L264-L269


JoshuaC215:
Nice catch, thanks!

PR #44, opened 2024-10-05
Add the "I" series rules in ruff (isort operation)

gbaian10:
Due to the mixing of built-in imports and third-party imports, it is difficult to distinguish the source of the packages.
Used the "I" series rules in ruff (an operation similar to isort) to solve this issue.

All the code in this PR was modified using `ruff check --fix`, and future checks can be done through pre-commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,5 +50,8 @@ dev = [
 line-length = 100
 target-version = "py310"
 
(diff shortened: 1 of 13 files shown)


JoshuaC215 (approved):

JoshuaC215:
This is great! Thank you!

PR #46, opened 2024-10-06
Add the "U" series rules in ruff (pyupgrade)

gbaian10:
Since both this project and ruff explicitly specify the minimum version as 3.10, it is hoped that syntax can be enhanced based on this version.

[documentation](https://docs.python.org/3/library/typing.html#deprecated-aliases)

[UP006](https://docs.astral.sh/ruff/rules/non-pep585-annotation/) deprecated in 3.9
[UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation/) support in 3.10
[UP035](https://docs.astral.sh/ruff/rules/deprecated-import/) deprecated in 3.9

All the code in this PR was modified using ruff check --fix, and future checks can be done through pre-commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,7 @@ line-length = 100
 target-version = "py310"
 
 [tool.ruff.lint]
-extend-select = ["I"]
+extend-select = ["I", "U"]
 
 [tool.pytest_env]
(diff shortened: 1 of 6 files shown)


JoshuaC215 (approved):
Awesome! 👏

PR #48, opened 2024-10-07
Complete type hints and Refactor some code

gbaian10:


src/schema/test_schema.py
```diff
@@ -38,15 +38,15 @@ def test_messages_from_langchain():
...
+def test_message_run_id_usage() -> None:
     run_id = "847c6285-8fc9-4560-a83f-4e6285809254"
     lc_message = AIMessage(content="Hello, world!")
     ai_message = ChatMessage.from_langchain(lc_message)
     ai_message.run_id = run_id
     assert ai_message.run_id == run_id
 
 
-def test_messages_tool_calls():
+def test_messages_tool_calls() -> None:
```
JoshuaC215, line 49: Is there a benefit to doing this for tests? It isn’t a common practice as far as I know

src/agent/research_assistant.py
```diff
@@ -18,7 +19,7 @@
 from agent.tools import calculator
 
 
-class AgentState(MessagesState):
+class AgentState(MessagesState, total=False):
```
JoshuaC215, line 22: What is this?

src/streamlit_app.py
```diff
@@ -103,27 +103,27 @@ def architecture_dialog():
...
         for m in messages:
             yield m
 
     await draw_messages(amessage_iter())
 
     # Generate new message if the user provided new input
-    if input := st.chat_input():
-        messages.append(ChatMessage(type="human", content=input))
-        st.chat_message("human").write(input)
+    if _input := st.chat_input():
```
JoshuaC215, line 113: Why add an underscore here?

src/schema/test_schema.py
gbaian10: The main issue is that ruff and mypy will complain if they don't have a return type annotation (if you have it set).

Of course, I agree that for tests, these return type hints aren't particularly useful. If you think it's better not to annotate in tests, I can change this part back.

src/streamlit_app.py
gbaian10: [ruff A001](https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/)

It's advisable to avoid using variable names that are the same as built-in variables, as this can sometimes lead to unexpected issues.
Of course, in this example, input() is not used, so there won't be any problems.

I personally think the simplest way to keep it straightforward is to add a leading underscore.
If you don't like underscores, you can change it to a more meaningful name, such as `user_input` or `chat_input`.

src/agent/research_assistant.py
gbaian10: [PEP589](https://peps.python.org/pep-0589/#totality)

If `total=False` is missing, the IDE or mypy will complain that you're missing some keys, as it assumes all keys are required by default.

image


Of course, this isn't a perfect change for the project, because you inherit from `MessagesState`, which is also a `TypedDict` but lacks `total=False`.
So in this case, `messages` will be considered required, while `safety` and `is_last_step` are optional.

If you consider `messages` to be optional as well, you can choose not to inherit from the prebuilt `MessagesState` and instead add it to your own `TypedDict`.

However, regarding `total=False`, whether it's set or not, it won't affect your runtime.
Python won't check or enforce the presence of these keys.
This feature is mainly for static checking and IDE assistance.

src/schema/test_schema.py
JoshuaC215: Ah OK I see even in Langchain itself they do this. I hadn't seen that before in a popular library (of course I don't always go spelunking through the tests)

https://github.com/langchain-ai/langchain/blob/master/libs/core/tests/unit_tests/messages/test_utils.py

I'm fine with it then. TY!

src/agent/research_assistant.py
JoshuaC215: Mind just adding a comment with the link so it isn't confusing?

https://peps.python.org/pep-0589/#totality

src/streamlit_app.py
JoshuaC215: Let's change it to `user_input` TY

JoshuaC215 (approved):
LGTM other than the couple nits in comments, will merge after those are resolved. Thank you!

src/streamlit_app.py
gbaian10: Finish

src/agent/research_assistant.py
gbaian10: According to the explanation at the top of the PEP, I changed it to the [official document link](https://typing.readthedocs.io/en/latest/spec/typeddict.html#totality).

PR #51, opened 2024-10-08
Update .gitignore

gbaian10:
Prevent JetBrains users from submitting their `.idea` folder (similar to the .vscode folder).


JoshuaC215:
I haven’t used JetBrains much so I’m not too familiar. I see the comment right above this change says that making it is not recommended and suggests a bigger template. Whats your experience with this? Is there a norm to ignore the whole directory?

gbaian10:
Yes.
I checked the [Langchain repo](https://github.com/langchain-ai/langchain/blob/master/.gitignore#L3), and they also ignore the entire folder.

PR #54, opened 2024-10-10
Enhance the API documentation.

gbaian10:
1. Add FeedbackResponse Response
1. Add SSE Response
image

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -155,7 +155,7 @@ def _sse_response_example() -> dict[str, Any]:
             "description": "Server Sent Event Response",
             "content": {
                 "text/event-stream": {
-                    "example": "data: {'type': 'token', 'content': 'Hello'}\n\ndata: {'type': 'token', 'content': ' World'}\n\n",
+                    "example": "data: {'type': 'token', 'content': 'Hello'}\n\ndata: {'type': 'token', 'content': ' World'}\n\ndata: [DONE]\n\n",
                     "schema": {"type": "string"},
                 }
             },
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -15,7 +15,14 @@ from langgraph.graph.state import CompiledStateGraph
 from langsmith import Client as LangsmithClient
 
 from agent import research_assistant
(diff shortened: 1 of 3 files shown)


src/service/service.py
```diff
@@ -142,7 +149,21 @@ async def message_generator(user_input: StreamInput) -> AsyncGenerator[str, None
...
 
 
-@app.post("/stream")
+def _sse_response_example() -> dict[str, Any]:
+    return {
+        status.HTTP_200_OK: {
+            "description": "Server Sent Event Response",
+            "content": {
+                "text/event-stream": {
+                    "example": "data: {'type': 'token', 'content': 'Hello'}\n\ndata: {'type': 'token', 'content': ' World'}\n\n",
```
JoshuaC215, line 158: nit: should we add `data: [DONE]\n\n` here for completeness / correctness? Otherwise this change looks good TY!

JoshuaC215 (approved):

src/service/service.py
gbaian10: I think a better approach is to use `event` to mark the DONE event, rather than using a specific string in the data.

At present, you might only need two events:  `data` and  `done`.

If you add the event, it also has a great visual effect in Postman.

https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events-intro

image

JoshuaC215: Sorry I’m not an expert in this and not sure I’m following the change you’re suggesting. Is it something to include in this PR or a follow on?

gbaian10: image

Just modify the `yield` part to use the function mentioned above.

gbaian10: > Sorry I’m not an expert in this and not sure I’m following the change you’re suggesting. Is it something to include in this PR or a follow on?

I think changes can be made in a future PR.

JoshuaC215: Ok - for now the example response is not an accurate example since it doesn’t include the current [DONE] string - mind adding it? And then fix to whatever in future PR

gbaian10: Updated!

PR #58, opened 2024-10-12
OpenAI model optional

gbaian10:
OpenAI model optional and Update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ With that said, there are several other interesting projects in this space that 
    ```sh
    # Provide at least one LLM API key to enable the agent service
 
-   # Optoinal, to enable OpenAI gpt-4o-mini
+   # Optional, to enable OpenAI gpt-4o-mini
    OPENAI_API_KEY=your_openai_api_key
 
    # Optional, to enable LlamaGuard and Llama 3.1
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ This project offers a template for you to easily build and run your own agents u
 
 [![Streamlit App](https://static.streamlit.io/badges/streamlit_badge_black_white.svg)](https://agent-service-toolkit.streamlit.app/)
 
-
 <a href="https://agent-service-toolkit.streamlit.app/"><img src="media/app_screenshot.png" width="600"></a>
 
 ### Quickstart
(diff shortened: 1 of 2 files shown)


gbaian10:
I also added a feature that allows for the deletion of inactive models when starting Streamlit.
But haven't added it to this PR yet.

If I only provide the OpenAI API key.

image

How do you think the results are?

```py
from agent.research_assistant import models as enabled_models

models = {
    "OpenAI GPT-4o-mini (streaming)": "gpt-4o-mini",
    "Gemini 1.5 Flash (streaming)": "gemini-1.5-flash",
    "Claude 3 Haiku (streaming)": "claude-3-haiku",
    "llama-3.1-70b on Groq": "llama-3.1-70b",
}

unenable_model = [k for k, v in models.items() if v not in enabled_models]
for k in unenable_model:
    del models[k]
```

README.md
```diff
@@ -80,19 +79,22 @@ With that said, there are several other interesting projects in this space that
...
 
 2. Set up environment variables:
    Create a `.env` file in the root directory and add the following:
-   ```
-   # OPENAI_API_KEY is the only required variable
+
+   ```sh
+   # Provide at least one LLM API key to enable the agent service
+
+   # Optoinal, to enable OpenAI gpt-4o-mini
```
JoshuaC215, line 97: typo: `Optoinal`

src/agent/research_assistant.py
```diff
@@ -31,10 +31,9 @@ class AgentState(MessagesState, total=False):
 
 # NOTE: models with streaming=True will send tokens as they are generated
 # if the /stream endpoint is called with stream_tokens=True (the default)
-models = {
-    "gpt-4o-mini": ChatOpenAI(model="gpt-4o-mini", temperature=0.5, streaming=True),
-}
-
+models: dict[str, str] = {}
```
JoshuaC215, line 34: wouldn't it be `dict[str, Any]` (`langchain.BaseChatModel` or whatever it is)

JoshuaC215:
> I also added a feature that allows for the deletion of inactive models when starting Streamlit.
But haven't added it to this PR yet.

It sounds convenient but I don't feel too strongly about adding it.

My suggestion: I still want to add the feature for supporting multiple agents which can be referenced by name (such as via a /path/ or body or header parameter). If we added that feature, we could have a general service metadata endpoint which returns the available agents and available models? And then the streamlit could be configured on that.

If you wanted to build it like that or in another way that would be easy to extend to also specify multiple agents, I'd be supportive. Thoughts?

README.md
gbaian10: fixed

src/agent/research_assistant.py
gbaian10: fixed

JoshuaC215 (approved):

Issue #52, opened 2024-10-08
Load previous session


antonioalegria:
Would be interesting to add loading the conversation history if re-using a thread ID from a previous session. It would basically require adding a service endpoint to get the history and then calling that in the beginning to populate it.

A separate issue would be to have the list of threads, like in ChatGPT.

JoshuaC215:
Cool idea! I think it would be a significant change and tough to generalize. Also won’t work with the demo app since I don’t have a persisted db running today. Open to a contribution here but would need to be pretty neatly abstracted and not add to core complexity. Not sure I will have time to work on it.

antonioalegria:
Sure. I was thinking adding a service endpoint that would get the history for a thread id, basically all of the messages and the current state, and then it would draw them in the app. This would be called first instead of drawing the placeholder first message.

JoshuaC215:
Ah nice - yeah maybe it would be straightforward - feel free to share a pull request :)

peterkeppert:
I already have this implemented for my project along the lines of @antonioalegria 's suggestion. So I did a quick port to agent service toolkit to see what would need to be done.

For now I added Chat ID to the sidebar, so that user can copy it and use it to share the chat like this: http://localhost:8502/?thread_id=c45ba586-31e3-4a22-9858-207f89830436 . Once such chat from history is opened, you can continue conversation in that thread.

![image](https://github.com/user-attachments/assets/801582dc-5c8b-4cda-ae78-e698b64b66c2)

antonioalegria:
Any reason to use Chat ID instead of Thread ID?

peterkeppert:
Not really, just overlooked this relic of my implementation, I updated the pull request to consistently use Thread ID

JoshuaC215:
Thanks again for adding this @peterkeppert and @antonioalegria for the suggestion!

PR #62, opened 2024-10-15
Chat History

peterkeppert:
Draft PR for #52 to get an idea what needs to be done for basic functionality.

### DIFF
--- a/src/streamlit_app.py
+++ b/src/streamlit_app.py
@@ -6,7 +6,7 @@ import streamlit as st
 from streamlit.runtime.scriptrunner import get_script_run_ctx
 
 from client import AgentClient
-from schema import ChatMessage
+from schema import ChatMessage, ChatHistory
 
 # A Streamlit app for interacting with the langgraph agent via a simple chat interface.
 # The app has three main functions which are all run async:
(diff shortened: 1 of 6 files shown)


src/streamlit_app.py
```diff
@@ -87,14 +99,14 @@ def architecture_dialog() -> None:
                 "Prompts, responses and feedback in this app are anonymously recorded and saved to LangSmith for product evaluation and improvement purposes only."
             )
 
+        st.markdown(f"Chat ID: **{st.session_state.thread_id}**")
```
JoshuaC215, line 102: Add something like `help="Set ?thread_id=… to retrieve this conversation"`  so it’s more obvious?

peterkeppert: Great idea, done!

peterkeppert:
Right, I will try to create a service test.

As for more work needed, I was primarily thinking about the UI side of things, that it looks rather rudimentary. And also that maybe it is not so clear when we are continuing an existing chat and when we are having a new one and might want to save the Thread ID before closing the window.

JoshuaC215:
Sweet - yeah from my side happy to merge after you add a test, and we can do further UI work in follow-on PRs. Whatever you prefer. (if the test gets too complex we could possibly delay that too)

peterkeppert:
Service test added, please check.

JoshuaC215 (approved):
Nice work! Looks good!

PR #63, opened 2024-10-19
feat: add HTTPBearer check

gbaian10:
The current API places Bearer authentication at the middleware layer, which prevents direct access to the API documentation from a browser when an API KEY is set.
Additionally, the need to provide this API KEY is not mentioned in the API documentation.

I used the built-in features of FastAPI to implement this functionality and removed the middleware that provided the same functionality.
Additionally, because I wanted to allow access to `docs`, `redoc`, and `openapi.json` without providing an API KEY, I set up an extra APIRouter.

docs:
image

redoc:
image

---

If you agree with this change, I can add the corresponding tests.
However, the existing tests will fail when an API KEY is provided, so the old tests will need to be adjusted accordingly.

I believe that as the codebase grows, this project needs some refactoring.
For example, tests should have a separate test folder, and routers should have their own modules.

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -1,13 +1,14 @@
 import json
 import os
 import warnings
-from collections.abc import AsyncGenerator, Callable
+from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
-from typing import Any
+from typing import Annotated, Any
 from uuid import uuid4
 
-from fastapi import FastAPI, HTTPException, Request, Response, status
+from fastapi import APIRouter, Depends, FastAPI, HTTPException, status
 from fastapi.responses import StreamingResponse
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from langchain_core._api import LangChainBetaWarning
 from langchain_core.runnables import RunnableConfig
 from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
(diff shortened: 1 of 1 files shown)


JoshuaC215:
Nice, makes sense and looks great! Thank you!

JoshuaC215:
For now I don’t think a router module is needed but agree it could make sense if / as more complexity is added.

Would be fine with consolidating tests to their own folder now especially with the change to src/ format

gbaian10:
Do I need to add tests to validate this feature?

Currently, the tests are executed without the `AUTH_SECRET` environment variable.
When your `.env` file includes this variable, the tests will fail.

I tried to check for the environment variable before the tests start and insert a fake one if it doesn't exist.
However, since `service.py` and `test_service.py` are in the same folder, and the `__init__.py` in that folder executes the following behavior:
```py
from service.service import app

__all__ = ["app"]
```

This causes the `app` object to be instantiated before entering the test file, which affects my intention to simply insert a fake `AUTH_SECRET` (Because it impacting `service.py`).

This is why I want to test independently in a separate folder.
If it doesn't need testing, then maybe I don't need to add anything extra, and it can be turned into a formal PR.

JoshuaC215:
1) I'd be OK to merge w/o new tests if we'd need a bigger refactor to make the testing work. I did local manual testing and LGTM!

2) I was not able to replicate the behavior of `.env` causing tests to fail, either on your branch or my current branch. The `.env` should not be loading by default in the case of running tests unless you added some tooling that loads it? However I can replicate the behavior if I run e.g. `AUTH_SECRET=12345 pytest` from cmdline, both on current main and your branch. So not obvious to me that this PR has any regression, but maybe I missed something?

3) It seems like you could check for the existence of the environment variable only when the route is called - e.g. remove the global setting `bearer_depend = [Depends(verify_bearer)] if os.getenv("AUTH_SECRET") else None` and only check that inside of verify_bearer? Then the import wouldn't matter and you could set the ENV only for the test. I guess another / additional approach would be a factory method that creates the app and allows you to override the ENV in testing? And we call that method with the ENV when running for real.

gbaian10:
Ready for the official PR!

WillamysC:
It is not clear to me how it works. Do I need to provide a key or it creates one?

JoshuaC215:
@WillamysC you need to set AUTH_SECRET environment variable when starting the server. It will use whatever value you set, does not auto generate. If not set, it runs unauthenticated.

Issue #65, opened 2024-10-24
Unexpected message type: <class 'schema.schema.ChatMessage'>


OzDuys:
When using the hosted app in the README this error is returned.

JoshuaC215:
Needed a reboot, should be fixed now. Thanks for reporting!

PR #70, opened 2024-10-26
Fix class type bug in Streamlit app (?)

JoshuaC215:
It's been common for a while to see a bug in the app deployed to Streamlit Cloud where it starts throwing errors related to class type mismatch after a code change. The issue is always resolved after restart.

Screenshot 2024-10-26 at 11 48 43 AM

This has been reported a few times e.g. https://github.com/JoshuaC215/agent-service-toolkit/issues/65 and https://github.com/JoshuaC215/agent-service-toolkit/discussions/47

I have a hypothesis that the issue occurred because the app was caching the AgentClient instance which then got out of sync with the new code somehow (e.g. some bytecode around the class type or something). This would only happen on Streamlit Cloud since it updates the app "live" and maintains the cache. I've never been able to repro it locally (you could probably do it with hot reloading).

This PR updates to keep the AgentClient in session_state instead of a cache, so it will reset more often. If my hypothesis is correct, then the bug might still occur if you have an open session during an app update but I think that would be very rare and acceptable.

Will monitor for a while after deploying this fix and see if the issue reoccurs.

### DIFF
--- a/src/streamlit_app.py
+++ b/src/streamlit_app.py
@@ -23,12 +23,6 @@ APP_TITLE = "Agent Service Toolkit"
 APP_ICON = "🧰"
 
 
-@st.cache_resource
-def get_agent_client() -> AgentClient:
-    agent_url = os.getenv("AGENT_URL", "http://localhost")
-    return AgentClient(agent_url)
-
-
 async def main() -> None:
     st.set_page_config(
         page_title=APP_TITLE,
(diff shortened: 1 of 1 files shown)


JoshuaC215:
Ok, the bug was occurring in the app BEFORE I merged this PR and then resolved after, without a restart. That's a good sign that my hypothesis was correct! 🎉

PR #77, opened 2024-10-31
Implement CustomData for agents; select agent in app

JoshuaC215:
- Add a CustomData class in agents/ for sending custom messages either as a node output or custom event
- Add a selectbox in the streamlit app for picking the agent to run


JoshuaC215:
cc @peterkeppert FYI

PR #78, opened 2024-11-01
fix bg tasks display logic

peterkeppert:
@JoshuaC215 , let's finish this in your branch. after your review, I have updated 3 things here:
In `task.py`: Task statuses so that the "started" entry gets displayed when task is started
In `streamlit_app.py`:
- Logic to determine container states based on child tasks' states
- Removed "break" after encountering an "error" task state. If there is another task besides the error one which is still running, we should wait until they all complete, before declaring the container state to be "error".


JoshuaC215:
@peterkeppert I pushed some bug fixes to your PR.

I think in your version you were trying to support multiple parallel bg tasks? But I still don't think the code worked the way you intended, since only one status container was ever open at a time. At least to me the logic and for loop seemed confusing.

If you want to make parallel bg tasks work in the UI before merging this PR, please make it so `bg_task_agent.py` sends multiple parallel bg tasks and test to make sure it works how you expect.

Otherwise if you're OK with this version we can merge it in and merge the branch to main.

peterkeppert:
Thanks for the bugfix! Initially, I didn't envision having parallel tasks, only multiple sequential tasks, e.g.: Generate retrieval prompts, Perform document retrieval, Generate response, Evaluate response. However, we can easily support parallel tasks as well. I have added demonstration to `bg_task_agent`, where Task 1 starts first and finishes last and Task 2 runs in between.

Having one status container for all tasks that run between receiving last Human Message and next AI Message was a decision we made early on. Having multiple status containers was taking up way too much space in the chat, so we decided to have one status container and details of task runs available within.

Also, I have made these changes:
1. Added methods completed() and completed_with_error() to TaskData - I didn't like comparing the field values in UI to determine the task state and it was prone to trivial bugs.
2. Added type hint to `Task` result field.
3. I reverted to the previous state where we initally set `st.status("")`. The logic has been updated that status label will only change if we encounter a new task. So when Task 1 runs, status container changes it's label to Task 1. Then it stays that way until Task 2 starts. This is the final label of the container no matter which task finishes first. Status of the container is `running` until all tasks complete. If all tasks complete succesfully, status changes to `complete`. If there was at least one error, status changes to `error` and user can inspect the container to see went wrong.

JoshuaC215:
I cleaned this up a bit and added better code comments. Some of the design choices don't make sense to me - I would probably show the first task as the label instead of the last new task since the first one is probably the parent task? But I'll merge it.

I'd like to factor out all this rendering logic somehow since it's now 20% of the streamlit app which isn't part of the main functionality. Maybe into a separate class which can accept new TaskData inputs and render itself in Streamlit. But it can be a separate PR. I dunno if you want to work on that or if I should.

peterkeppert:
I think which Task to show as container label depends on the use case - for sequential messages case which I outlined (and that is the primary one I have in mind), it makes sense to show last started Task, so you can track execution progress without expanding the status container.

For parallel tasks, it depends on their relationship. They might be parent/child tasks or completely independent tasks which are just all needed for the app to generate a response. In the first case, showing first Task makes sense, in the second case, I think both options are reasonable and it’s hard to generalize. I would leave this to the implementators to customize based on the use case they want to cover.

As for extracting the gui part to a separate class, I would prefer to leave this one to you.

Issue #57, opened 2024-10-12
Model Name Error


gbaian10:
The current model accepts any string as input.
If an incorrect model name is entered or a model is provided without an API key, a 500 error will occur.

![image](https://github.com/user-attachments/assets/60d91b42-b38e-4476-9c0f-f966710200ee)

It is recommended to use an Enum or dynamic Enum to check these values, allowing Pydantic to catch them and raise a 422 error
Or to define a custom error.

JoshuaC215:
[Sounded like](https://github.com/JoshuaC215/agent-service-toolkit/discussions/64#discussioncomment-11041772) you're working on a draft PR which would address this using Enums. Looking forward to it!

gbaian10:
Hi, would you consider raising the project's minimum Python version to 3.11?
I would like to use many practical features of 3.11, but I'm unable to do so due to the project's minimum version requirement.

For example, [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum) has many advantages over the regular Enum (Enum is indeed a very useful feature).
Python 3.11 also has significant performance improvements over 3.10.

Additionally, this project supports Langgraph Cloud, but [Langgraph's Docker Image](https://hub.docker.com/r/langchain/langgraph-api/tags) does not support 3.10.

This project uses uv for project management, and uv seems to have quite convenient features for switching versions.

Additionally, this project is a template, not a package, so I believe upgrading the version usually won't affect too many users (users won't import modules from this project as a third-party package).
I believe encouraging the use of newer versions of Python is also a good direction.

Another, it's a bit strange that the CI only tests against Python 3.12.
Usually, tests should be conducted for all supported versions, or at least the minimum version.
Otherwise, if I use a new syntax introduced in Python 3.12, it will pass in CI, but users with 3.10 or 3.11 will encounter issues.

JoshuaC215:
Sure, we can update to 3.11. Good point about fixing the CI to test the minimum version.

gbaian10:
> [Sounded like](https://github.com/JoshuaC215/agent-service-toolkit/discussions/64#discussioncomment-11041772) you're working on a draft PR which would address this using Enums. Looking forward to it!

I recently started switching computers, so there might be some delays in progress.

PR #86, opened 2024-11-10
add pydantic-settings and ModelEnum

gbaian10:
## pydantic-settings
- https://github.com/JoshuaC215/agent-service-toolkit/discussions/64#discussion-7352693

## model_Enum
- #57

The minimum required version of Python has been upgraded to `3.11`.

---

image

I haven't yet confirmed whether the LLM is functioning properly on AWS and Google.

The external enum names have been adjusted. Perhaps you prefer the original names?
My goal is to minimize hard-to-remember inputs, such as the date in "claude-3-haiku-20240307" and the suffix in "llama-3.1-70b-versatile."
The external API is kept as simple as possible, but internally, these names are mapped to their original versions through the `_MODEL_TABLE` and are centrally managed in `core/llm.py`, allowing for easy changes and refactoring.

This PR includes many adjustments but no functional changes; it's mostly refactoring.
There should be plenty to discuss, so I've set this PR as a draft.


src/client/client.py
```diff
@@ -13,7 +13,7 @@ class AgentClient:
 
     def __init__(
         self,
-        base_url: str = "http://localhost:80",
+        base_url: str = settings.BASE_URL,
```
JoshuaC215, line 16: I’m torn on whether to use the settings in client vs sticking with os for simplicity. I want to add a metadata endpoint so it can get the model names from the server instead of hard coding it.

If we add it to client like this, you’ll need to add pydantic-settings to the top level requirements.txt or this will break the hosted streamlit app

gbaian10: I forgot that the client side is separate, because I was experimenting in the same environment. 😅️
In this case, to indeed reduce external dependencies, it's better not to install `pydantic-settings` and directly use `os`.

gbaian10: update

src/agents/llama_guard.py
```diff
@@ -76,13 +77,11 @@ def parse_llama_guard_output(output: str) -> LlamaGuardOutput:
...
     def __init__(self) -> None:
-        if os.getenv("GROQ_API_KEY") is None:
+        if settings.GROQ_API_KEY is None:
             print("GROQ_API_KEY not set, skipping LlamaGuard")
             self.model = None
             return
-        self.model = ChatGroq(model="llama-guard-3-8b", temperature=0.0).with_config(
-            tags=["llama_guard"],
-        )
+        self.model = get_model(GroqModelName.LlAMA_GUARD_3_8B)
```
JoshuaC215, line 84: I don't think we should hide the `with_config` inside of get_model, it's too easy to miss for this case. Can we keep it explicit in this file instead?

src/core/llm.py
```diff
@@ -0,0 +1,89 @@
...
+    AWSModelName.BEDROCK_HAIKU: "anthropic.claude-3-5-haiku-20241022-v1:0",
+}
+
+AllModelEnum: TypeAlias = (
+    OpenAIModelName | AnthropicModelName | GoogleModelName | GroqModelName | AWSModelName
+)
+ModelT: TypeAlias = ChatOpenAI | ChatAnthropic | ChatGoogleGenerativeAI | ChatGroq | ChatBedrock
+
+
+@cache
```
JoshuaC215, line 71: Is this optimization doing anything useful here? I'm slightly nervous about side effects or weirdness for negligible benefit... maybe it's fine though

src/core/config.py
```diff
@@ -0,0 +1,71 @@
...
+
+    OPENAI_API_KEY: SecretStr | None = None
+    ANTHROPIC_API_KEY: SecretStr | None = None
+    GOOGLE_API_KEY: SecretStr | None = None
+    GROQ_API_KEY: SecretStr | None = None
+
+    OPENAI_DEFAULT_MODEL: OpenAIModelName = OpenAIModelName.GPT_4O_MINI
+    ANTHROPIC_DEFAULT_MODEL: AnthropicModelName = AnthropicModelName.HAIKU_3
+    GOOGLE_DEFAULT_MODEL: GoogleModelName = GoogleModelName.GEMINI_15_FLASH
+    GROQ_DEFAULT_MODEL: GroqModelName = GroqModelName.LLAMA_31_70B
```
JoshuaC215, line 37: Is the idea here that the developer could change the default model by setting this environment variable?

I think it might be overkill, and a bit neater to just have one `DEFAULT_MODEL = OpenAIModelName.GPT_40_MINI` defined in `core.llm` instead.

Then the model creation in agents would be simpler:

```python
from core.llm import get_model, DEFAULT_MODEL
# ...
get_model(config["configurable"].get("model", DEFAULT_MODEL))
```

JoshuaC215: Alternatively, if you want to make this configurable, can we instead do a simpler `settings.DEFAULT_MODEL` ? And you could check that it's valid for the API key, and if it isn't set in `model_post_init()` you could check which API keys have been set and pick a default model based on that.

OK, I talked myself into maybe it's a separate PR - if you want to defer I can file an issue

src/core/llm.py
```diff
@@ -0,0 +1,89 @@
...
+def get_model(model_name: str) -> ModelT:
+    # NOTE: models with streaming=True will send tokens as they are generated
+    # if the /stream endpoint is called with stream_tokens=True (the default)
+    api_model_name = _MODEL_TABLE.get(model_name)
+
+    if model_name in OpenAIModelName._value2member_map_:
```
JoshuaC215, line 77: Is checking `_value2member_map_` really the canonical way to determine if a string is in an enum? 😮

src/run_client.py
```diff
@@ -1,43 +1,46 @@
...
-for message in client.stream("Share a quick fun fact?"):
-    if isinstance(message, str):
-        print(message, flush=True, end="|")
-    elif isinstance(message, ChatMessage):
-        print("\n", flush=True)
-        message.pretty_print()
-    else:
-        print(f"ERROR: Unknown type - {type(message)}")
+if __name__ == "__main__":
+    asyncio.run(amain())
```
JoshuaC215, line 46: need to run `main()` after this, right?

src/streamlit_app.py
```diff
@@ -49,8 +55,7 @@ async def main() -> None:
         st.rerun()
 
     if "agent_client" not in st.session_state:
-        agent_url = os.getenv("AGENT_URL", "http://localhost")
-        st.session_state.agent_client = AgentClient(agent_url)
+        st.session_state.agent_client = AgentClient()
```
JoshuaC215, line 58: I think the `os.getenv` check is still needed, no? Maybe a missed revert when removing settings from the client side

JoshuaC215:
Looks good except those comments. Once those are addressed I'll pull it down locally to play around a bit and make sure it looks good on my side too, and then happy to merge. Awesome contribution!! Thank you!!

src/run_client.py
gbaian10: It's not necessary; I just want to say that this file looks like a test entry file.
There shouldn't be any files importing this module, right?

src/core/llm.py
gbaian10: I think writing it directly as `if model_name in OpenAIModelName` should work.I tested it just now and there were no issues.

Initially, I wanted to use match to write this part, but couldn't find a suitable way.
Eventually, I succeeded in making the match syntax work by using `_value2member_map_`.

However, I found it too strange in the end, and using if is simpler.

src/core/llm.py
JoshuaC215: Oh sorry I was referring to the use of @cache

src/run_client.py
JoshuaC215: Correct it’s just for manual testing the API without Streamlit. not for importing

src/core/llm.py
gbaian10: I used the @cache method to ensure that when the same model_name is input, the same object can be retrieved without having to recreate it.
In fact, these BaseChatModels seem to require creating an httpx.client internally, which makes object creation take quite a bit of time. On a better computer, it might take only 0.05 to 0.3 seconds, while on a less powerful one, it could take even longer.

Essentially, this cache is just an alternative to using a global dict to store models. I think that writing it as a function is easier to read than directly accessing objects from a dict using a key.

gbaian10:
Do we need a separate `.env.example` file to make it easier for users to create a `.env` file?
I once considered doing this, but then I found similar instructions in the `README.md`. However, having a standalone `.env.example` file might also be a good choice.

src/core/llm.py
JoshuaC215: Sounds good, I didn't realize about the httpx client. Good catch! 👍

JoshuaC215:
`.env.example` sounds great. Maybe just update the README to have a minimal example and link to the new file for a full reference. Might want to add that as a comment in the settings file too so it's more memorable to update.

src/agents/llama_guard.py
gbaian10: fixed

src/core/config.py
gbaian10: fixed

src/core/llm.py
gbaian10: fixed

src/streamlit_app.py
gbaian10: fixed

gbaian10:
I've been quite busy recently and haven't updated for a while.
I've almost completed all the requirements mentioned above, and everything runs smoothly locally.
However, I'm currently unsure why it fails in the CI's Docker test. It seems unable to find the `core` module.

JoshuaC215:
@gbaian10 OK if I work on cleaning up the remaining things and then I can merge?

JoshuaC215:
One thing to fix is that schema/ now depends on core/ which imports dotenv, langchain and initializes settings, and client/ depends on /schema, so all of that is required when you run the client separately (part of why the docker is breaking).

I think I will move the LLM definitions into schema but keep get_model in core. Then I think the separation will be clean again.

gbaian10:
> @gbaian10 OK if I work on cleaning up the remaining things and then I can merge?

Yes, you can first test to see if it works as expected, then make some code changes before merging.
I haven't tested Google and AWS, and the Enum is also missing links to their model cards.

gbaian10:
Additionally, I'm not sure which of `streamlit_app.py`, `run_client.py`, and `client/client.py` are specific to the client side (and must be used with only the packages installed via `requirements.txt`).
Have you considered moving these client-only files into the `client` folder, and possibly moving `requirements.txt` there as well?

gbaian10:
image

What should the minimum version for this project be: 3.10, 3.11, or 3.12?

JoshuaC215:
OK I will finish cleaning it up tomorrow.

Re: separating out the client, it could be a future PR. To do it cleanly would mean moving to 3 separate packages: the schema, the agent + service, and the client + app. I'm not sure if it's worth so much complexity at this point but something to consider.

PR #99, opened 2024-11-26
Another attempt on Python 3.13 testing in CI

JoshuaC215:


JoshuaC215:
Looks like [langchain-google-genai v2.0.11](https://github.com/langchain-ai/langchain-google/tree/libs/genai/v2.0.11) has the needed upgrades, I will take another crack at this very soon

PR #124, opened 2024-12-27
Add Chat Sharing Dialog & BugFix Empty History

fedebotu:
This PR adds the following improvements:

## 1. Add Chat Sharing Dialog

Previously, a user had to manually copy the `thread_id` to the service URL and set it as a query parameter. Now, a new dialog appears that allows users to conveniently copy the full link for sharing/resuming chats:

share-resume-chat

## 2. BugFix Empty History

If the user tried for any reason to open a new chat with a `thread_id` with no messages yet, the below error appeared:

server-error-no-history

This PR fixes this issue by simply checking whether there is any history. If there is not, a simple error popup appears, avoiding the crash:

history-fix

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ classifiers = [
 requires-python = ">=3.11, <3.13"
 
 dependencies = [
-    "duckduckgo-search>=6.4.1",
+    "duckduckgo-search>=7.1.1",
     "fastapi ~=0.115.5",
     "httpx ~=0.27.2",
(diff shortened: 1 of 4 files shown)


JoshuaC215:
Hey @fedebotu thank you for this awesome contribution!!

I was trying to apply a small patch and screwed up my local git commands which caused this to be closed in a way I can't seem to revert 😿 I'm sorry for that!

I applied the fix in #125 and ensured your commits would stay in tact so you show as an author.

PR #150, opened 2025-01-23
Update message_generator for multi-agent workflows

TaurusLondon:
When building multi-agent workflows, an event can be a Command to navigate among agents and it will break the message_generator method.

May need to use the latest version of langgraph (e.g. 0.2.66)


JoshuaC215:
Hey @tauruslondon thanks for raising this, I didn’t see the Command launch until now.

Can you describe in more detail and/or give some example agent code where using Commands breaks the message_generator logic? It wasn’t immediately obvious to me, so it would help with validating this PR.

Thank you!

JoshuaC215:
I did a more robust fix in https://github.com/JoshuaC215/agent-service-toolkit/pull/159

PR #152, opened 2025-01-25
Add devcontainer config

vargasmontero:


JoshuaC215:
Hey, thanks for opening this! Looks like a non customized devcontainer config which means it has a couple of problems:
1. This repo doesn't use `pip install -r requirements.txt`, it uses `uv` (see the README)
2. The app requires the service to be running in order to function, so the service would also need to be started before running the app

If you want to tweak it to actually work with the repo it'd be welcome! Otherwise I'll take a look when I have time or may close this. Cheers!

JoshuaC215:
Happy to re-open if there's more movement

PR #164, opened 2025-02-02
[DRAFT] Add Experimental Model Context Protocol (MCP) Support

madtank:
# [DRAFT] Add Experimental Model Context Protocol (MCP) Support

## Overview
Introduces an optional, lightweight integration of Model Context Protocol (MCP) capabilities into the agent service toolkit, enabling advanced file and context management through a flexible, opt-in approach.

## Key Changes
- Add conditional Node.js and MCP server installation in Docker container
- Implement file system operations with restricted access via agent_home
- Add environment-based feature flags for MCP functionality
- Create flexible, low-overhead mechanism for experimental features

### Core Changes
- Modified `docker/Dockerfile.service` for conditional MCP installation
- Added `agent_home` volume in compose.yaml for persistent storage
- Updated `.env.example` with MCP configuration options

### New Features
- File system operations with controlled access
- Tool-based file management interface
- Persistent storage through Docker volumes

## Configuration
Enable MCP features by setting in .env:
```env
MCP_ENABLED=true   # Enable MCP features
AGENT_HOME=/app/agent_home  # File system workspace
```

## Benefits
- Zero-cost abstraction when disabled
- Containerized installation
- Restricted file system access
- Easy to enable/disable
- Supports future extensibility

## Current Status
- Basic file operations working
- Docker integration complete
- Environment configuration ready
- Ready for initial feedback

Testing
Manual testing completed:

Container builds with and without MCP enabled
File operations working in restricted directory
Tool integration with agent framework
Volume persistence across restarts

Example Usage
Here are screenshots demonstrating the core functionality:

File Operations Demo:

Listing allowed directories
Creating a sample file
Reading file contents
Verifying file persistence

Directory Restrictions:

Proper isolation to /app/agent_home
File creation and modification
Content updates maintaining restrictions

agent-mcp-file-edits

## After creating the file in the agent_home directory, the agent can now read the file in a new session
 
agent-mcp-file-read

## Breaking Changes
None - Feature is opt-in and doesn't affect existing functionality

## Future Work (Separate PRs)
- Enhanced documentation
- Additional MCP capabilities
- Comprehensive testing suite
- Dependency management improvements

## Notes
- Draft PR for initial feedback
- Feature is experimental and opt-in
- Focus on minimal, secure implementation
- Looking for feedback on approach and implementation

## Related Issues


docker/Dockerfile.service
madtank, line 1: This file is WIP, will work on minimal changes.

src/agents/research_assistant.py
madtank, line 1: We might need new agent, consider human in the loop for some actions like delete file.

JoshuaC215: +1 prefer to have a different agent.

madtank: Agreed, since this is running in Docker, next MCP I'd add is shell, so human in the loop is good idea.

JoshuaC215:
Thanks for getting this started! Couple of things I want to flag:

- I would STRONGLY prefer to use the [HTTP SSE transport procotol](https://modelcontextprotocol.io/docs/concepts/transports#server-sent-events-sse) and run the server on a separate docker image, instead of stdio and baking it into the existing service image as you have in this PR. I believe this will be much more consistent with the existing architecture of the repo, have a nicer de-coupling, and be easier to maintain. I think it's also more consistent with how someone would run MCP in a production environment rather than a toy example.
- Ideally any MCP server implementation lives in its own directory of `src/`, not inside `src/agents/`. Actually even better would be to take an MCP server off the shelf such as an existing docker image vs re-implementing here but I dunno if that's feasible. Adding a lightweight client in the src/agents/tools seems reasonable.
- I don't want to add file manipulation tools into this template repo. File manipulation driven by an LLM on the host system is dangerous and easy to abuse. We should pick a different example which is safe.

JoshuaC215:
BTW if you are committed to the approach of using stdio for local agent and file manipulation / shell, it might make sense to do this as a forked project. I'm happy to promote and support it if you decide to go with that approach!

madtank:
Thank you for the detailed feedback! I completely agree with your points about the architecture and security considerations. You're absolutely right that using SSE transport with a separate Docker container would be more production-ready and better aligned with the repo's architecture.

I'm happy to pivot away from file manipulation tools - security is paramount. Instead, I can look into integrating some of the safer, officially supported MCP servers. For example:
- The Brave Search server for web queries
- Simple calculation/analytics tools
- Memory/knowledge graph capabilities

Or let me know if you prefer one of the ones from the list here:
https://github.com/modelcontextprotocol/servers?tab=readme-ov-file

Would you have any specific suggestions for example tools that would better demonstrate MCP integration while maintaining security? I'm thinking we could make the tools configurable, allowing users to opt-in to specific capabilities while keeping the template focused on safe defaults.

I'll work on restructuring the implementation to:
- Use SSE transport instead of stdio
- Run MCP server(s) in separate container(s)
- Move the client code out of src/agents/
- Focus on safer example tools

Let me know if you have any thoughts on which direction would be most valuable for the template repo.

JoshuaC215:
Here are a few that look like good options to me, one of these or something similar:
- [knowledge graph memory](https://github.com/modelcontextprotocol/servers/tree/main/src/memory)
- [fetch a url](https://github.com/modelcontextprotocol/servers/tree/main/src/fetch)
- [code sandbox](https://github.com/Automata-Labs-team/code-sandbox-mcp/tree/main)
- [local duckdb](https://github.com/motherduckdb/mcp-server-motherduck) - would need to include a sample data file to use with it

it seems like a lot of the reference servers use the stdio transport instead of sse. I don't really understand how that works well between docker containers, but if there's a way to make that work with one of the existing provided containers (don't need to copy everything over) that would be OK too.

I found a typescript implementation of the SSE server: https://github.com/modelcontextprotocol/servers/blob/main/src/everything/sse.ts

JoshuaC215:
btw, most of your notes seem to follow a similar conversation cadence to an LLM, I wondered if you're using it to support your writing or what's going on with that? (it's ok with me but was just curious)

madtank:
> Here are a few that look like good options to me, one of these or something similar:
> 
> * [knowledge graph memory](https://github.com/modelcontextprotocol/servers/tree/main/src/memory)
> * [fetch a url](https://github.com/modelcontextprotocol/servers/tree/main/src/fetch)
> * [code sandbox](https://github.com/Automata-Labs-team/code-sandbox-mcp/tree/main)
> * [local duckdb](https://github.com/motherduckdb/mcp-server-motherduck) - would need to include a sample data file to use with it
> 
> it seems like a lot of the reference servers use the stdio transport instead of sse. I don't really understand how that works well between docker containers, but if there's a way to make that work with one of the existing provided containers (don't need to copy everything over) that would be OK too.
> 
> I found a typescript implementation of the SSE server: https://github.com/modelcontextprotocol/servers/blob/main/src/everything/sse.ts

Yes, these are good MCP servers, I think the first two are good options. I was thinking knowledge graph memory would be interesting, and would want to consider the implecation in an enterprise setting, would users all have the same knowledge graph in this case because it would be a shared server, first thing that comes to mind that could be pretty interesting if the agent is setup to have a persona and this would give it the ability to adapt, then I think of data poisoning, there will always be that risk of abuse. I'm good with whatever, my thinking is needs to be modular and something each user would want to swap out the tools.

madtank:
> btw, most of your notes seem to follow a similar conversation cadence to an LLM, I wondered if you're using it to support your writing or what's going on with that? (it's ok with me but was just curious)

Yes, I mostly use Claude Desktop app with MCP of course. It's great for planning and brainstorming.

Here's the updated plan.
image

JoshuaC215:
Nice :) Plan looks good. As you said any tool should be fine as it shows how to set it up and user can swap their own.

Memory is a good one which should be easy to modularize and demonstrate. Let's go with that!

I think the single shared memory is fine for this example, it can be called out in the documentation. There's a similar issue today where I can access another users chat history on the same server if I guess their thread_id (by default a UUID so not a substantial real risk).

madtank:
> Nice :) Plan looks good. As you said any tool should be fine as it shows how to set it up and user can swap their own.
> 
> 
> 
> Memory is a good one which should be easy to modularize and demonstrate. Let's go with that!
> 
> 
> 
> I think the single shared memory is fine for this example, it can be called out in the documentation. There's a similar issue today where I can access another users chat history on the same server if I guess their thread_id (by default a UUID so not a substantial real risk).

Yeah, shared memory could be interesting if used correctly and understood. You can tell it to save something and then in a later session you can have it retrieve, true for a single user or many users on the same server. Seems like a good way to share information. Maybe with quantum computing UUID could be a risk, besides that would be brute force or with right access. 

I appreciate the feedback hopefully mcp adds some value. From my experience it works well with this project, blazing fast 🚀

madtank:
@JoshuaC215 After working with MCP for a while and thinking through the integration, I agree that it makes sense to keep this fork as more of a personal project for now. I’ll explore cleaner ways to integrate it, possibly through native LangChain or another library.

For now, I’ve merged changes to my main branch with MCP running in a Docker container, including shell and file access. I’ve found that it likely benefits from a multi-agent setup, and it’s easy to hit rate limits quickly.

I’ll leave the draft open if you’d like to revisit it, but for now, I don’t have the bandwidth to focus on it. It works well, but getting it to an enterprise-ready state would require more time and understanding than I can commit to at the moment. I’ll circle back if a strong use case emerges.

JoshuaC215:
Sounds good 🫡 thanks for the update and your work on this!!

reatang:
This is a huge but rewarding job.

It is best to make it an independent agent `mcp_assistant.py`

madtank:
This langchain mcp adaptor looks like a good fit. Still need to do a bit more research, but might have something. I'll update how it goes.

reatang:
Button `New Chat` ！
It's a feature I really want! But I can't write `streamlit`

madtank:
> Button `New Chat` ！
> 
> It's a feature I really want! But I can't write `streamlit`

Look at my fork and see how I added. I have new chat and quick start questions.

madtank:
@JoshuaC215 I've made good progress on integrating MCP into my forked main branch. I'm able to easily switch out tools; memory, fileservice, and brave search all work well in my testing. The integration with langchain-mcp-adapters is very clean. However, what I've found is most MCP servers do not natively work with SSE—they only support stdio. So in my repo, I'm installing npm on the service container and running MCP servers there. However, this will work well with SSE if the MCP server supports it, or if we use a proxy. I'll spend some time looking into a proxy. For now, I'm happy with the direction it's going.

JoshuaC215:
I'm going to close this for inactivity. I'm still supportive of this change, feel free to re-open if/when you want to work on it again!

madtank:
I'm happy with the way that I was able to integrate MCP into my fork repo. But I agreed with some of your concerns about putting this in your repo. If you ever want to check out how I did it look at the MCP agent and if you're interested, I can do something similar. It was very easy using langchains adapter and overall pretty clean. I turned my repo into a primary MCP agent. It's pretty cool, but not something you'd want to put in a production enterprise system. Most of all there is just risk with excessive agency.

PR #169, opened 2025-02-10
Postgresql checkpointer support.

scchengaiah:
Hello @JoshuaC215 ,

Thank you for this implementation. Was really helpful to get started faster.

Thought adding a Postgresql checkpointer to the current implementation might be useful for enterprise use cases when we deploy this solution on cloud.

Could you please review and share your thoughts ?

Regards,
Chengaiah

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -28,12 +28,12 @@ LANGCHAIN_API_KEY=
 # Application mode. If the value is "dev", it will enable uvicorn reload
 MODE=
 
-# Database type. 
+# Database type.
 # If the value is "postgres", then it will require Postgresql related environment variables.
-# If the value is "sqllite", then you can configure optional file path via SQLITE_DB_PATH
+# If the value is "sqlite", then you can configure optional file path via SQLITE_DB_PATH
 DATABASE_TYPE=
 
-# If DATABASE_TYPE=sqllite (Optional)
+# If DATABASE_TYPE=sqlite (Optional)
 SQLITE_DB_PATH=
 
 # If DATABASE_TYPE=postgres
--- a/.env.example
+++ b/.env.example
@@ -28,6 +28,21 @@ LANGCHAIN_API_KEY=
 # Application mode. If the value is "dev", it will enable uvicorn reload
 MODE=
 
+# Database type. 
+# If the value is "postgres", then it will require Postgresql related environment variables.
+# If the value is "sqllite", then you can configure optional file path via SQLITE_DB_PATH
+DATABASE_TYPE=
+
+# If DATABASE_TYPE=sqllite (Optional)
+SQLITE_DB_PATH=
+
+# If DATABASE_TYPE=postgres
+POSTGRES_USER=
+POSTGRES_PASSWORD=
+POSTGRES_HOST=
+POSTGRES_PORT=
+POSTGRES_DB=
+
 # OpenWeatherMap API key
 OPENWEATHERMAP_API_KEY=
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,7 @@ dependencies = [
     "langgraph-checkpoint-postgres ~=2.0.13",
     "langsmith ~=0.1.145",
     "numexpr ~=2.10.1",
+    "psycopg[binary,pool] ~=3.2.4",
     "pyarrow >=18.1.0", # python 3.13 support
     "pydantic ~=2.10.1",
     "pydantic-settings ~=2.6.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,7 @@ dependencies = [
     "langchain-ollama ~=0.2.3",
     "langgraph ~=0.2.68",
     "langgraph-checkpoint-sqlite ~=2.0.1",
+    "langgraph-checkpoint-postgres ~=2.0.13",
     "langsmith ~=0.1.145",
     "numexpr ~=2.10.1",
     "pyarrow >=18.1.0", # python 3.13 support
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -1,7 +1,14 @@
 from typing import Annotated, Any
 
 from dotenv import find_dotenv
-from pydantic import BeforeValidator, HttpUrl, SecretStr, TypeAdapter, computed_field
+from pydantic import (
+    BeforeValidator,
+    Field,
+    HttpUrl,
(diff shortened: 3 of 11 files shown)


pyproject.toml
```diff
@@ -29,6 +29,7 @@ dependencies = [
     "langchain-ollama ~=0.2.3",
     "langgraph ~=0.2.68",
     "langgraph-checkpoint-sqlite ~=2.0.1",
+    "langgraph-checkpoint-postgres ~=2.0.13",
     "langsmith ~=0.1.145",
     "numexpr ~=2.10.1",
```
jo3p, line 34:
```suggestion
    "numexpr ~=2.10.1",
    "psycopg[binary,pool] ~=3.2.4",
```

jo3p (changes requested):
Please also run `ruff format src/` to format using the same linter as configured in `pyproject.toml`. It looks like you used a different formatter which makes it harder to distinguish actual code changes from cosmetic changes by the linter.

.env.example
```diff
@@ -28,6 +28,21 @@ LANGCHAIN_API_KEY=
 # Application mode. If the value is "dev", it will enable uvicorn reload
 MODE=
 
+# Database type. 
+# If the value is "postgres", then it will require Postgresql related environment variables.
+# If the value is "sqllite", then you can configure optional file path via SQLITE_DB_PATH
+DATABASE_TYPE=
+
+# If DATABASE_TYPE=sqllite (Optional)
```
JoshuaC215, line 36: nit: typo `sqlite`

src/core/settings.py
```diff
@@ -61,6 +68,26 @@ class Settings(BaseSettings):
     )
     LANGCHAIN_API_KEY: SecretStr | None = None
 
+    # Database Configuration
+    DATABASE_TYPE: str = "sqlite"  # Options: "sqlite" or "postgres"
```
JoshuaC215, line 72: Possible to define this as an enum with fixed values similar to the models instead of str? (e.g. AllModelEnum)

I would be fine to just do it in this same file

src/run_service.py
```diff
@@ -6,4 +9,9 @@
 load_dotenv()
 
 if __name__ == "__main__":
-    uvicorn.run("service:app", host=settings.HOST, port=settings.PORT, reload=settings.is_dev())
+    # Set Compatible event loop policy on Windows Systems.
+    if sys.platform == "win32":
+        asyncio.set_event_loop_policy(asyncio.WindowsSelectorEventLoopPolicy())
```
JoshuaC215, line 14: Is this required for this PR or an unrelated change? For the unfamiliar reader (like me) could you link to some docs on this in the comment or otherwise give a little more detail on why it's needed?

src/service/service.py
```diff
@@ -39,10 +41,45 @@
...
+
 def verify_bearer(
     http_auth: Annotated[
         HTTPAuthorizationCredentials | None,
-        Depends(HTTPBearer(description="Please provide AUTH_SECRET api key.", auto_error=False)),
+        Depends(
+            HTTPBearer(
+                description="Please provide AUTH_SECRET api key.", auto_error=False
+            )
+        ),
```
JoshuaC215, line 82: Yeah +1 the other comment, this looks like you're using a formatter other than the one configured in the repo. Please use the pre-commit checks in the repo (see Contributing section of README) otherwise CI will fail.

src/service/service.py
```diff
@@ -39,10 +41,45 @@
...
+
+
+def get_postgres_connection_string() -> str:
+    """Build and return the PostgreSQL connection string from settings."""
+    return (
+        f"postgresql://{settings.POSTGRES_USER}:"
+        f"{settings.POSTGRES_PASSWORD.get_secret_value()}@"
+        f"{settings.POSTGRES_HOST}:{settings.POSTGRES_PORT}/"
+        f"{settings.POSTGRES_DB}"
+    )
```
JoshuaC215, line 72: can we move these to a separate file like `src/service/postgres.py` and import here?

src/service/service.py
```diff
@@ -53,19 +90,56 @@ def verify_bearer(
...
+        }
+
+        try:
+            async with AsyncConnectionPool(
+                conninfo=get_postgres_connection_string(),
+                max_size=settings.POSTGRES_POOL_SIZE,
+                min_size=settings.POSTGRES_MIN_SIZE,
+                max_idle=settings.POSTGRES_MAX_IDLE,
+                kwargs=connection_params,
+            ) as pool:
```
JoshuaC215, line 113: Can we make the pool contructor into a separate function and put it in `src/service/postgres.py` with the functions above? And then call it here, so there's not much postgres-specific logic here.

JoshuaC215:
@scchengaiah I left a few comments but this is awesome, thanks so much for the contribution.

.env.example
scchengaiah: Done

scchengaiah:
@JoshuaC215 : Applied the suggested modifications, Could you please validate and share your thoughts ?

pyproject.toml
scchengaiah: Done

src/core/settings.py
scchengaiah: Done

src/run_service.py
scchengaiah: Added elaborated documentation.

src/service/service.py
scchengaiah: Done

src/service/service.py
scchengaiah: Done, created a module named memory and moved the implementation.

src/service/service.py
scchengaiah: Done, created a module named memory and moved the implementation.

JoshuaC215:
Need to add `COPY src/memory/ ./memory/` to docker/Dockerfile.service.

Also would be nice to add memory to the watch paths for the service in compose.yaml (actually I just noticed core is missing from the watch paths too). I can do that in a separate PR if you aren’t familiar.

scchengaiah:
> Need to add `COPY src/memory/ ./memory/` to docker/Dockerfile.service.
> 
> Also would be nice to add memory to the watch paths for the service in compose.yaml (actually I just noticed core is missing from the watch paths too). I can do that in a separate PR if you aren’t familiar.

I will try to incorporate these changes in sometime

Issue #141, opened 2025-01-17
Consider adding an template for a tool node that implements RAG solution


gplutov:

JoshuaC215:
I would welcome a PR for this

See discussion in https://github.com/JoshuaC215/agent-service-toolkit/discussions/142

JoshuaC215:
This has been completed with #171

PR #171, opened 2025-02-18
Issue #141 - Add RAG assistant template

tpetursdottir:
- Add simple helper to create chroma db with PDF or Word Documents.
- Create a RAG assistant, with customised instructions on what its speciality is and what information it has access to in the database.
- Add database_search_func to the tools, this could also be imported in the research_assistant and added to the list of tools there.
- Customise AI greeting message to fit the RAG agent's speciality and describe what it has knowledge of.


JoshuaC215:
Very nice!!! Would you want to add a brief README section describing what the new agent is and how to use it?

Also, thoughts on including a few files for a usable sample implementation? Could be a later PR as well

helpers/create_chroma_db.py
```diff
@@ -0,0 +1,84 @@
+import os
+import shutil
```
JoshuaC215, line 2: Maybe call this folder scripts/ instead?

src/agents/rag_assistant.py
```diff
@@ -0,0 +1,142 @@
+from datetime import datetime
```
JoshuaC215, line 1: Looks like CI is failing on linter in this file, please run pre-commit

helpers/create_chroma_db.py
tpetursdottir: Updated to scripts

src/agents/rag_assistant.py
tpetursdottir: CI failures should hopefully be fixed now :)

src/agents/tools.py
```diff
@@ -38,3 +40,31 @@ def calculator_func(expression: str) -> str:
 
 calculator: BaseTool = tool(calculator_func)
 calculator.name = "Calculator"
+
+
+# Create the embedding function for our project description database
+embeddings = OpenAIEmbeddings()
```
JoshuaC215, line 46: Looks like initializing this as a global breaks the service when OpenAI API key environment variable is missing (a common configuration)

src/agents/tools.py
```diff
@@ -38,3 +40,31 @@ def calculator_func(expression: str) -> str:
...
+
+
+# Format retrieved documents
+def format_contexts(docs):
+    return "\n\n".join(doc.page_content for doc in docs)
+
+
+# Load the stored vector database
+chroma_db = Chroma(persist_directory="./chroma_db", embedding_function=embeddings)
```
JoshuaC215, line 55: Will this cause a problem if chroma db was not set up? Would also prefer if running the service without RAG does not create an extra empty db file

JoshuaC215: It may be fine to just move all of this inside the tool function, not sure there’s a high cost to repeatedly re-initializing? Or set it up only the first time it’s called

src/agents/tools.py
tpetursdottir: Moved inside a function.

pyproject.toml
```diff
@@ -30,13 +30,14 @@ dependencies = [
...
+    "langchain-chroma ~=0.2.2",
     "langgraph-checkpoint-sqlite ~=2.0.1",
     "langgraph-checkpoint-postgres ~=2.0.13",
     "langgraph-supervisor ~=0.0.8",
     "langsmith ~=0.1.145",
     "numexpr ~=2.10.1",
-    "numpy ~=1.26.4; python_version <= '3.12'",
-    "numpy ~=2.2.3; python_version >= '3.13'",
+    "numpy ~=1.26.2; python_version <= '3.12'",
+    "numpy ~=1.26.2; python_version >= '3.13'",
```
JoshuaC215, line 40: This conflicts with some later changes, I was trying locally to rebase and get uv lock to pass and having a very bad time.

It seems like LangChain team recently patched langchain-chroma to support later numpy versions (https://github.com/langchain-ai/langchain/pull/30714) but they haven't released it yet. I tried installing from git but uv was giving me a bad time.

Seems like we need to either wait for langchain-chroma 0.2.3 to be released with the latest update (preferred) or otherwise install from source to fix the conflict.

PR #177, opened 2025-02-28
Add openai compatible

reatang:
Add an OpenAI-compatible configuration to allow new models (such as local LM-Studio) to be quickly added for testing

### DIFF
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -113,7 +113,7 @@ class Settings(BaseSettings):
     def model_post_init(self, __context: Any) -> None:
         api_keys = {
             Provider.OPENAI: self.OPENAI_API_KEY,
-            Provider.OPENAI_COMPATIBLE: self.COMPATIBLE_API_KEY,
+            Provider.OPENAI_COMPATIBLE: self.COMPATIBLE_BASE_URL,
             Provider.DEEPSEEK: self.DEEPSEEK_API_KEY,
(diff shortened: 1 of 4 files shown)


JoshuaC215:
Hey, can you explain a little more on how this would be configured and used?

reatang:
If I want to quickly test an unconnected LLM, I can directly configure
COMPATIBLE_MODEL=
COMPATIBLE_API_KEY=
COMPATIBLE_BASE_URL=
and select `openai-compatible` on the UI

src/core/settings.py
```diff
@@ -126,6 +133,10 @@ def model_post_init(self, __context: Any) -> None:
                     if self.DEFAULT_MODEL is None:
                         self.DEFAULT_MODEL = OpenAIModelName.GPT_4O_MINI
                     self.AVAILABLE_MODELS.update(set(OpenAIModelName))
+                case Provider.OPENAI_COMPATIBLE:
+                    if self.DEFAULT_MODEL is None:
+                        self.DEFAULT_MODEL = OpenAICompatibleName.OPENAI_COMPATIBLE
+                    self.AVAILABLE_MODELS.update(set(OpenAICompatibleName))
```
JoshuaC215, line 139: Should we add some lightweight validation checks here that the needed settings were set?

reatang: You are right, it has been added

JoshuaC215 (approved):

Issue #173, opened 2025-02-20
support langgraph-supervisor


xytian315:
Langgraph developed a new library https://github.com/langchain-ai/langgraph-supervisor for mutli-agent implementation.

Is there any plan to support this? 

Currently service.py streamlit UI will break if using langgraph-supervisor library to implement an agent

JoshuaC215:
Hey, thanks for reporting. I just heard about this supervisor library from your issue.

At a quick glance it looks like something we should support similar to Command. I’m not sure about the effort required.

Can you share the error you see and some more details like any logs? It will help inform the fix.

I would also welcome a PR if anyone else sees this and knows what is needed.

xytian315:
Hi @JoshuaC215, 

Sorry for late reply! thanks for getting back to me. 

The Errors Regarding service

```
    |   File "/github/agent-service-toolkit/src/service/service.py", line 160, in message_generator
    |     new_messages = event["data"]["output"].update.get("messages", [])
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | AttributeError: 'list' object has no attribute 'get'
```

Additionally, issues in Streamlit.app, specifically the section ```# Expect one ToolMessage for each tool call.```

and ```Unexpected ChatMessage type: tool```

and ```Error parsing message: Unsupported message type: tuple```

JoshuaC215:
Thanks! I'll take a closer look when I have some time (probably next 1-2 weeks?), unless someone else writes a PR first.

PR #178, opened 2025-02-28
Feat: convert to astream() and add langgraph-supervisor library support

xytian315:
Related with https://github.com/JoshuaC215/agent-service-toolkit/issues/173

**This PR updates to Langgraph v0.3 and converts the main streaming api from `astream_events()` to the better supported [astream()](https://langchain-ai.github.io/langgraph/how-tos/streaming/).** 

Hi @JoshuaC215 

Could you please take a look at this draft MR? It is not entirely ideal, but would love to know your suggestions how to solve this problem. 

1. When supervisor and agent produce the same AI message, what do we do about it? Right now displaying both, what is your opinion on this one? 
For example, a simple greeting message, both Agent and Supervisor message will produce the same AI message content, should we do filter out any duplicate message in this case?

(I believe this scenario is also the same if we just use command_agent without using langgraph-supervisor lib)

2. In streamlit.app, right now I am skipping all the handoff messages https://github.com/langchain-ai/langgraph-supervisor-py/blob/main/langgraph_supervisor/handoff.py. And I am skipping all message type.tool(as they seem to dup as ai.tool_calls messages)

Looking forward to your advice.


JoshuaC215:
Hey, nice! Yeah I bet we can add some simple filter in the streaming to prevent the duplicate messages (probably close to where you already modified service.py). Broadly, this makes sense. It will be good to include a sample agent that uses supervisor in the final PR, or modify an existing agent to use it.

I think I will have time in the next 2-3 days to look more closely at this and help hack on it, or certainly within the next week. Sorry for the delay!

xytian315:
Thanks for reply! I just did a commit. I used the sample agent from repo https://github.com/langchain-ai/langgraph-supervisor-py

JoshuaC215:
Hey, I spent some time messing with this tonight. The duplicated messages seem to happen because every subsequent agent is streaming its own events. I didn't find a way to track how deep in the hierarchy a given agent is automatically. I think the filter would need to be aware of which agents it's supposed to be streaming from.

Also looks like LangChain changed the API again >_< and `graph.stream()` is the preferred way instead of `stream_events()`. IDK if that's also impacting things.

If you want to get this working I'd suggest hooking up your calls to langsmith so you can trace through all the calls + metadata and figuring out what filters you need. It seems to be pretty complex. I think the tool call filter you did in the streamlit app could also be moved to the service (I'd request making that change before merging)

I'm not sure I'll have time in the next few weeks to go through this and figure it out but I welcome it if you're able to.

xytian315:
@JoshuaC215  Thank you so much for taking a look and providing suggestions. I will take a look. I really appreciate your help here.

xytian315:
FYI: @JoshuaC215  
I played around ```graph.stream()``` a little bit. It definitely made things much easier

graph.stream() changed a lot of stuff, especially tool messages stuff. I think we should change this soon in this project. Do you have any plan to change to ```graph.stream()``` soon? 
 
 (I am currently playing around it, there are some uncertainties in terms of UI behavior, not sure how to handle it)

JoshuaC215:
Nice, thanks for playing with this and getting it started! Welcome any more commits in this PR if you make further progress.

Sounds like it makes sense to assess a migration, I don't think it existed when I first built this, or at least didn't work as well. I _think_ I will also have some time this weekend to look more closely at what you did and see about migrating.

xytian315:
@JoshuaC215  Thanks so much for taking a look!

I tested the change in this branch with the existing agents including chat, assistant, Command, bg-task-agent, they all seem working well. 

only langgraph-supervisor agent using the lib got some weird behavior, I created an issue here https://github.com/langchain-ai/langgraph-supervisor-py/issues/60

JoshuaC215:
So much good stuff in this PR!! Really happy we could migrate to the latest Langgraph stremaing API. Thank you @xytian315 !!! 👏 👏 👏 👏  🎉 🎉 🎉 🎉 

I did some cleanups:
- Added back support for token by token streaming with the new API
- Added back support for custom events with the new API (slight update to the interface cc @peterkeppert)
- I added another slight shim in the service to fix the supervisor output so it matches the usual expected format, so the streamlit app doesn't have to change (felt cleaner)

PR #180, opened 2025-03-07
Interrupt handling

jleonelion:
[interrupts are the recommended method](https://langchain-ai.github.io/langgraph/concepts/human_in_the_loop/#interrupt) for human in the loop behavior (data validation, confirming action to take, etc.).  I've enhanced this codebase to support basic interrupts (where the inputs and outputs are simple strings).  More advance interrupts handling (ex. interrupt value or response is JSON payload) requires further customization, but I'm trying to keep this simple so somebody can expand if/when they need to.

### DIFF
--- a/src/agents/agents.py
+++ b/src/agents/agents.py
@@ -5,9 +5,9 @@ from langgraph.graph.state import CompiledStateGraph
 from agents.bg_task_agent.bg_task_agent import bg_task_agent
 from agents.chatbot import chatbot
 from agents.command_agent import command_agent
+from agents.interrupt_agent import interrupt_agent
 from agents.langgraph_supervisor_agent import langgraph_supervisor_agent
 from agents.research_assistant import research_assistant
-from agents.interrupt_agent import interrupt_agent
 from schema import AgentInfo
 
 DEFAULT_AGENT = "research-assistant"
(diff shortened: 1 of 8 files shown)


JoshuaC215:
Very nice! This looks promising, thanks for putting it together @jleonelion 

Looks like the formatting needs to be fixed by running pre-commit hook. I will take a closer look when I have some time and may push some changes to your branch if it's OK.

jleonelion:
Of course!  Apologies about the formatting. I must not have initialized the hooks…but I did make sure all the unit tests still passed.

iPhone. iTypo. iApologize.
________________________________

JoshuaC215:
@jleonelion this was awesome, I love the example interrupt agent with the zodiac sign check and it works great.

I recently switched from using `astream_events()` to `astream()` API and it simplifies the interrupt tracking quite a bit, I updated the PR to match. Also did a bit of other refactoring.

I found that with `ainvoke()` you could also set a stream mode to get it to return the interrupt directly, instead of needing to query agent state for it 🙌 .

I'm going to go ahead and merge this but feel free to leave any other comments or thoughts if you review and we could put in follow-on PRs.

Cheers and thank you for the stellar contribution!!

jleonelion:
@JoshuaC215 I'm glad this helped.  Keep up the good work.  I've starred your GitHub account now 😄

PR #183, opened 2025-03-10
improve google llms config

mab-happtiq:


JoshuaC215:
Nice, if you want to extend this to the other model types in the same style and maintain tests and formatting pass, I'm happy to merge it. Agree it simplifies the maintenance a bit and is more DRY.

PR #196, opened 2025-04-04
Langgraph store implementation

aeronesto:
* Significantly updated interrupt_agent with node that retrieves from and updates the LangGraph store. 
* Updated a node in the interrupt agent to allow more generalized conversations; it can have a conversation instead of just responding with his zodiac sign. 
* Implemented long-term memory (cross-thread memory) with LangGraph store.
* Created user ID cookie in browser storage to track returned users (for long-term memory).
* Added a Postgres service and volume for Postgres data to compose configuration.
* Updated documentation to reflect changes

### DIFF
--- a/src/agents/interrupt_agent.py
+++ b/src/agents/interrupt_agent.py
@@ -1,5 +1,6 @@
 from datetime import datetime
-from typing import Any
+from typing import Any, cast
+import logging
 
 from langchain.prompts import SystemMessagePromptTemplate
 from langchain_core.language_models.base import LanguageModelInput
@@ -14,6 +15,9 @@ from langgraph.types import interrupt
 from pydantic import BaseModel, Field
 
 from core import get_model, settings
+
+# Added logger
+logger = logging.getLogger(__name__)
 
 
 class AgentState(MessagesState, total=False):
@@ -37,7 +41,7 @@ def wrap_model(
 
 background_prompt = SystemMessagePromptTemplate.from_template("""
 You are a helpful assistant that tells users there zodiac sign.
-Provide a one paragraph summary of the origin of zodiac signs.
+Provide a one sentence summary of the origin of zodiac signs.
 Don't tell the user what their sign is, you are just demonstrating your knowledge on the topic.
 """)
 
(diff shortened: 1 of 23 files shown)


src/service/service.py
```diff
@@ -320,6 +334,7 @@ def history(input: ChatHistoryInput) -> ChatHistory:
             config=RunnableConfig(
                 configurable={
                     "thread_id": input.thread_id,
+                    "user_id": input.user_id,
```
JoshuaC215, line 337: Does adding this to `agent.get_state()` actually do anything? Is user_id needed in the history param? Since it's still just returning the messages for one thread...

src/agents/interrupt_agent.py
```diff
@@ -68,43 +74,140 @@ class BirthdateExtraction(BaseModel):
...
+                }
+        except Exception as e:
+            # Log the error or handle cases where the store might be unavailable
+            print(f"Error reading from store for namespace {namespace}, key {key}: {e}")
+            # Proceed with extraction if read fails
+            pass
+    else:
+        # If no user_id, we cannot reliably store/retrieve user-specific data.
+        # Consider logging this situation.
+        print("Warning: user_id not found in config. Skipping persistent birthdate storage/retrieval for this run.")
```
JoshuaC215, line 120: should not switch between print and logging

src/agents/interrupt_agent.py
```diff
@@ -68,43 +74,140 @@ class BirthdateExtraction(BaseModel):
...
+#     if not state.get("birthdate"):
+#         raise ValueError("No birthdate found in state")
+# 
+#     m = get_model(config["configurable"].get("model", settings.DEFAULT_MODEL))
+#     model_runnable = wrap_model(
+#         m, sign_prompt.format(birthdate=state["birthdate"].strftime("%Y-%m-%d"))
+#     )
+#     response = await model_runnable.ainvoke(state, config)
+# 
+#     return {"messages": [AIMessage(content=response.content)]}
```
JoshuaC215, line 178: Clean this up please and don't leave commented code (similar below)

src/memory/__init__.py
```diff
@@ -16,4 +16,11 @@ def initialize_database() -> BaseCheckpointSaver:
         return get_sqlite_saver()
 
 
-__all__ = ["initialize_database"]
+def initialize_store():
+    """
+    Initialize the appropriate store based on configuration.
+    Returns an async context manager for the initialized store.
+    """
+    return get_postgres_store()
```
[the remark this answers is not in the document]
aeronesto: LangGraph only supports a postgres store. There isn't one for SQL.

src/service/service.py
aeronesto: Good point, only the thread_id is needed.

src/memory/__init__.py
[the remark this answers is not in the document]
JoshuaC215: Sure, it should do the in memory store by default then unless the postgres type is set. otherwise it breaks the basic non-docker setup

aeronesto:
Just implemented your feedback and pushed the latest changes.

JoshuaC215:
Need to run pre-commit and fix the lint and format checks, then ensure tests pass

xytian315:
Question regarding this statement: 
> Created user ID cookie in browser storage to track returned users (for long-term memory).

It seems not quite accurate? Based on the code it seems like storing the user id in the stream lit session state, which is temporary, lives in the server side. It is not storing the user IDs in browser cookies on client side. 

( the reason why I am asking is that I was looking for the features to store user auth token etc in browser cookies. it doesn't seems like this MR introduced that feature)

JoshuaC215:
Yeah you're correct @xytian315, it does NOT use a cookie only exposes a URL param.

Streamlit [has native support for reading cookies](https://docs.streamlit.io/develop/api-reference/caching-and-state/st.context) but not for writing them. There are a few custom components to do it, [this one is the most popular](https://github.com/Mohamed-512/Extra-Streamlit-Components) although I'm not sure it works perfectly.

PR #200, opened 2025-04-10
Upgrade all dependencies

maver1ck:
This PR upgrades dependencies to newest versions.

uv lock was upgraded with `uv lock --upgrade`


JoshuaC215:
Not sure why it's needed but can you run the pre-commit lint check?

maver1ck:
There was issue in old code. Fixed.

Issue #199, opened 2025-04-08
Support for Mypy


maver1ck:
Hi @JoshuaC215,
I really appreciate your work. And I'd like to ask if you consider supporting mypy in the repo ?

Currently it's giving some error:
```
$ mypy .
...
Found 88 errors in 17 files (checked 31 source files)
```
I may help with fixing them :)

What do you think ?

Regards,
Maciek

JoshuaC215:
Hey @maver1ck, cheers, thanks for taking a look.

Yeah, I'm wondering how invasive the changes are but in principle I'm open to it. In particular, if it has a bad time with Langchain / Langgraph and requires a lot of extra boilerplate or something, I might want to live without it.

If it's not too bad and we want to add it, would also be nice to update the pre-commit and CI setup to check it going forward.

maver1ck:
Hi @JoshuaC215 
Can we start with some upgrades of packages / uv.lock ? Probably as a separate PR ?

For instance support for mypy in langgraph-supervisor is (3 weeks old)
Reference: 
https://github.com/langchain-ai/langgraph-supervisor-py/tree/main/langgraph_supervisor

maver1ck:
I will try do some fixes module after module.
Some problems may be non-trivial to fix because of mixed usage of AnyMessage and BaseMessage within Langchain and Langgraph. First PR created.

maver1ck:
@JoshuaC215 PR is ready to review

PR #201, opened 2025-04-10
Fix typing in memory module

maver1ck:
First part of fixes for mypy.
Fixes partially: https://github.com/JoshuaC215/agent-service-toolkit/issues/199

Before:
```
Found 88 errors in 17 files (checked 31 source files)
```

After:
```
Found 84 errors in 16 files (checked 31 source files)
```

### DIFF
--- a/src/memory/__init__.py
+++ b/src/memory/__init__.py
@@ -1,11 +1,14 @@
-from langgraph.checkpoint.base import BaseCheckpointSaver
+from contextlib import AbstractAsyncContextManager
+
+from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
+from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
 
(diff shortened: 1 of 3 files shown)


src/memory/__init__.py
```diff
@@ -1,11 +1,14 @@
...
+from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
+from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
 
 from core.settings import DatabaseType, settings
 from memory.postgres import get_postgres_saver
 from memory.sqlite import get_sqlite_saver
 
 
-def initialize_database() -> BaseCheckpointSaver:
+def initialize_database() -> AbstractAsyncContextManager[AsyncSqliteSaver | AsyncPostgresSaver]:
```
maver1ck, line 11: We can't use BaseCheckpointSaver here as it doesn't expose `setup()` method

PR #202, opened 2025-04-10
Add customizable PostgreSQL connection pool settings

JoshuaC215:
- Create explicit connection pool with configurable settings
- Use settings for min_size, max_size, and max_idle
- Update documentation with examples
- Add to README feature list

🤖 Generated with [Claude Code](https://claude.ai/code)
Co-Authored-By: Claude <noreply@anthropic.com>


src/memory/postgres.py
```diff
@@ -39,7 +40,33 @@ def get_postgres_connection_string() -> str:
...
+    """Initialize and return a PostgreSQL saver instance with connection pool."""
     validate_postgres_config()
-    return AsyncPostgresSaver.from_conn_string(get_postgres_connection_string())
+
+    # Create connection pool with custom settings
+    pool = create_connection_pool()
+
+    # Initialize saver with the pool
+    saver = AsyncPostgresSaver(conn=pool)
+    return saver
```
maver1ck, line 72: Hi @JoshuaC215,
Are you sure this would work ? Previously we were returning AsyncContextManager.

We're using this as context later on:
```
        async with initialize_database() as saver:
            await saver.setup()
```

JoshuaC215: Yeah fair enough, reverted in #209 after messing around with it for a bit. Going to just stick with the langgraph default setup.

PR #205, opened 2025-04-15
Fixing mypy

### LINKED ISSUE #199: Support for Mypy

maver1ck:
Fixes #199 

Done:
- all mypy errors with exception for changes from https://github.com/JoshuaC215/agent-service-toolkit/pull/202
- ignore of streamlit_app.py
- adding mypy to Github action

Result:
```
❯ mypy src  
src/memory/postgres.py:37: error: Item "None" of "SecretStr | None" has no attribute "get_secret_value"  [union-attr]
src/memory/postgres.py:71: error: Argument "conn" to "AsyncPostgresSaver" has incompatible type "AsyncConnectionPool[AsyncConnection[tuple[Any, ...]]]"; expected "AsyncConnection[dict[str, Any]] | AsyncConnectionPool[AsyncConnection[dict[str, Any]]]"  [arg-type]
src/memory/postgres.py:72: error: Incompatible return value type (got "AsyncPostgresSaver", expected "AbstractAsyncContextManager[AsyncPostgresSaver, bool | None]")  [return-value]
Found 3 errors in 1 file (checked 30 source files)
```


src/agents/chatbot.py
```diff
@@ -16,10 +15,10 @@ async def chatbot(
 ):
     messages = inputs["messages"]
     if previous:
-        messages = add_messages(previous["messages"], messages)
+        messages = previous["messages"] + messages
```
JoshuaC215, line 18: This (and below) slightly changes the behavior and deviates from the langgraph best practice as I understand it - is it needed?

Function definition is here, for reference: https://github.com/langchain-ai/langgraph/blob/main/libs/langgraph/langgraph/graph/message.py

JoshuaC215:
@maver1ck it's working now for me with a rebase. I would suggest to add mypy to the dev dependencies so it's always available instead of just installing in the CI.

See example passing pipelines and a couple additional commits in #210 (I tried to push to your branch but was unable to)

Note: it makes the non-docker CI stages 4-5x slower 😢  I wish mypy wasn't so slow, not sure how to address that. But they are still faster than the docker stage so I guess it's OK. I wonder if it should be after the test step instead of before, since the tests run much faster 🤔

src/agents/chatbot.py
maver1ck: This change fixes those errors which I found super difficult to fix any other way.
```
src/agents/chatbot.py:19: error: Incompatible types in assignment (expression has type "list[BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]] | BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]", variable has type "list[BaseMessage]")  [assignment]
src/agents/chatbot.py:19: error: Argument 1 to "add_messages" has incompatible type "list[BaseMessage]"; expected "list[BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]] | BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]"  [arg-type]
src/agents/chatbot.py:19: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
src/agents/chatbot.py:19: note: Consider using "Sequence" instead, which is covariant
src/agents/chatbot.py:19: error: Argument 2 to "add_messages" has incompatible type "list[BaseMessage]"; expected "list[BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]] | BaseMessage | list[str] | tuple[str, str] | str | dict[str, Any]"  [arg-type]
```

maver1ck:
Hi @JoshuaC215 
I rebased this PR, added mypy to dev deps. 

Regarding mypy speed it is possible to use some kind of cache. But I'm not sure how to set this up.

JoshuaC215:
Ya it seems like the mypy cache or other approaches are a big pain in the neck to set up 🤷

I also briefly tried setting up and running pyright which is supposed to be faster, but it spit out a bunch of new / different errors so I gave up.

Overall this is great!!! Thanks so much for pushing it through!

JoshuaC215 (approved):

PR #217, opened 2025-04-30
Support MongoDB checkpointer for persistence

alex-feel:
- Introduced langgraph-checkpoint-mongodb as a dependency
- Added MongoDB option in initialize_database function
- Implemented MongoDB configuration validation
- Created mongodb.py for MongoDB connection handling
- Updated DatabaseType to include MONGO


alex-feel:
Hmm, [this](https://github.com/JoshuaC215/agent-service-toolkit/actions/runs/14748064839/job/41399112987?pr=217) is strange since the pre-commit runs locally without any issues:

<details>

<summary>pre-commit run result</summary>



</details>

JoshuaC215:
Thanks for contributing, awesome! I want to test this on local before merging and I'm traveling for the next few days so it will be a bit before I can take a close review.

re: the uv format failure in CI, I'm not sure either. Maybe related to which version is installed somehow? Might be an older version in CI or something like that. The CI part is [defined here](https://github.com/JoshuaC215/agent-service-toolkit/blob/main/.github/workflows/test.yml#L26-L38). I can take a look if you don't figure it out before.

PR #226, opened 2025-05-26
Add Langfuse Tracing Support (Optional via LANGFUSE_TRACING)

mjeskulke:
Hi, thanks for the great repo — really appreciate the clean architecture!

This PR adds optional support for [Langfuse](https://www.langfuse.com/) tracing. Since we use Langfuse internally, this integration allows us to trace all agent calls when the LANGFUSE_TRACING environment variable is set to true.

In addition:
	•	A Langfuse check has been added to the /health endpoint.
To preserve existing behavior, the endpoint still returns healthy regardless of Langfuse status, but includes an additional langfuse field in the response payload when enabled.

Let me know if you’d like changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -72,3 +72,8 @@ OPENWEATHERMAP_API_KEY=
 
 # Agent URL: used in Streamlit app - if not set, defaults to http://{HOST}:{PORT}
 # AGENT_URL=http://0.0.0.0:8080
+
+# LANGFUSE Configuration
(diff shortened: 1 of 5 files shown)


JoshuaC215:
Very nice, thank you @mjeskulke !

I was going to make quick lint fixes but realized that I would be unable to push to your org fork. So, I hope you don't mind, I moved the commits to #229 to go ahead and merge and save you some time.

Issue #227, opened 2025-05-27
Should we try to design a new API interface to support the ag-ui protocol to interact with the front end?


reatang:
From the information I got, ag-ui should be the first standardized protocol for interacting with the front-end. I think we should study writing an example that supports this protocol in this project?

a12591771:
Very supportive

arjun-mavonic:
This would be great. Are we merging the PR anytime soon?

reatang:
> Thanks for suggesting it and for putting the PR together, and to everyone who left comments and 👍 . Sorry for my delay in review, I had a lot going on recently and didn't have as much time to dig into this.
> 
> I just reviewed and left a few modest feedbacks in the PR. I think we can get it merged before too long.

The review is just in time. I have been busy with the company's annual version launch this month. Today I have time to deal with it.

JoshuaC215:
Hey @reatang I recently came back to this and found that the ecosystem support has improved a lot, so we can include this in the toolkit without having to re-write a ton of code, in particular using https://pypi.org/project/ag-ui-langgraph/

Fix is now merged #323, let me know how it is if you try it out. Thank you for all your work on this and collaboration last year!!

PR #230, opened 2025-06-12
feat. Support AG UI protocol

reatang:
Although the ag-ui output library has been added, I think this protocol is not a mature protocol and even lacks many functions.

Completed issue：#227

### DIFF
--- a/docker/Dockerfile.service
+++ b/docker/Dockerfile.service
@@ -16,5 +16,6 @@ COPY src/memory/ ./memory/
 COPY src/schema/ ./schema/
 COPY src/service/ ./service/
 COPY src/run_service.py .
+COPY web-agui/ ./web-agui/
 
 CMD ["python", "run_service.py"]


src/service/service.py
```diff
@@ -305,6 +315,129 @@ async def message_generator(
         yield "data: [DONE]\n\n"
 
 
+async def message_generator_agui(
```
JoshuaC215, line 318: It seems like 80-90% of the logic in this method is the same as in the original, the main difference is the SSE encoding vs ag-ui encoding and adding the start and stop messages at the beginning.

Given that, won't it be much easier to maintain if we just add the protocol enum as an argument to the original message_generator function and do a switch or if/then for only the relevant parts? Otherwise it will be a pain to keep any logic changes up to date. It will also make this PR easier to review.

What do you think? Is there a reason to split it like this, or more differences that I missed?

JoshuaC215:
Looks good besides the comments. The protocol seems a little immature and I'm not sure about the long term maintenance (seems like maybe a startup created it?). But, it's not a huge addition after refactors so I don't think it's a high cost to have it in this project, and several people upvoted in the issue.

src/service/service.py
reatang: 1. I implemented AG-UI and hope everyone can discuss whether this protocol is easy to use
2. Because I think this protocol is still relatively simple

src/core/llm.py
```diff
@@ -104,6 +106,15 @@ def get_model(model_name: AllModelEnum, /) -> ModelT:
             openai_api_base="https://api.deepseek.com",
             openai_api_key=settings.DEEPSEEK_API_KEY,
         )
+    if model_name in AlibabaQWenModelName:
+        return ChatOpenAI(
```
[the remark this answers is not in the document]
reatang: OK, This is better

reatang:
> Looks good besides the comments. The protocol seems a little immature and I'm not sure about the long term maintenance (seems like maybe a startup created it?). But, it's not a huge addition after refactors so I don't think it's a high cost to have it in this project, and several people upvoted in the issue.

This protocol is a protocol layer hatched from the CopilotKit Agent front-end library. In theory, it can be directly used in CopilotKit, but I have not studied the CopilotKit front-end library in depth.

jump to：https://github.com/CopilotKit/CopilotKit

src/service/service.py
JoshuaC215: For sure, but I think duplicating the core message generating logic into two functions which must be kept up to date is a recipe for maintenance headaches later (or the agui one becoming stale)

src/schema/schema.py
```diff
@@ -75,6 +75,11 @@ class StreamInput(UserInput):
         default=True,
     )
 
+    stream_protocol: Literal["sse", "agui"] = Field(
```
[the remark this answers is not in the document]
JoshuaC215: Sounds good. I spent some time clicking through the copilotkit and ag-ui docs and still having trouble seeing how to launch a frontend that would interact with this server. Maybe checking in your pure frontend page and instructions to run it would be a good step? Unless you see a better way. I saw they have the "ag-ui dojo" but it's still not clear to me how to use it for a custom app.

reatang:
I have provided a single page to run a simple example, but this front-end page will not be a production-level feature.

src/service/service.py
reatang: From the perspective of a Technical Architect, I can learn from the design of the AG-UI
protocol, but I will not use an incomplete function in the products I provide to customers.
So we should still distinguish between SSE and AG-UI outputs.

src/service/service.py
```diff
@@ -404,4 +570,12 @@ async def health_check():
     return health_status
 
 
+@app.get("/agui/")
+async def agui_root():
+    with open("web-agui/index.html", encoding="utf-8") as file:
```
JoshuaC215, line 575: Just noting that this only works currently if the service is run from the root of the repo, and does not work if the service is in docker. If we move to a CLI or a website that just be launched directly from browser (no server interaction) we could just remove this.

reatang: This is indeed a difficult decision. If you want the agent to be a pure API service, then the web-agui function should also be separated into a rendering service like `streamlit_app`.

src/service/service.py
reatang: Great 👍

PR #233, opened 2025-06-18
Update VertexAI.md

nageCasillas:
docs: add note to COPY service-account-key.json in Dockerfile for VertexAI Docker support

### DIFF
--- a/docs/VertexAI.md
+++ b/docs/VertexAI.md
@@ -29,6 +29,12 @@ To use Vertex AI programmatically, you’ll create a **service account** and use
 #### 3. Add the JSON Key File to Your Project
 - Place the downloaded JSON file in the **root directory** of your project (e.g., `/my-project/service-account-key.json`).
 
+** If using Docker, also update `Dockerfile.service` to include:**
+```dockerfile
+COPY service-account-key.json service-account-key.json
+```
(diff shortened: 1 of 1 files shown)


docs/VertexAI.md
```diff
@@ -29,6 +29,12 @@ To use Vertex AI programmatically, you’ll create a **service account** and use
 #### 3. Add the JSON Key File to Your Project
 - Place the downloaded JSON file in the **root directory** of your project (e.g., `/my-project/service-account-key.json`).
 
+** If using Docker, also update `Dockerfile.service` to include:**
+```dockerfile
+COPY service-account-key.json service-account-key.json
```
JoshuaC215, line 34: Hey, thanks for flagging this and sorry the original instructions didn't include it.

Directly copying a sensitive credential into a docker image like this is dangerous, since the image might be copied or uploaded somewhere accidentally, leaking your credential.

A better approach would be to mount the file as a temporary volume, or to use a `docker secret`.

Want to update this to one of those approaches? And then will be happy to merge. Thanks for the contribution!

JoshuaC215: Yep the volume approach sounds fine. Thanks!

JoshuaC215 (approved):

PR #244, opened 2025-07-17
Adds sub graph/agent output to streamlit frontend 

haarisedhi102:
use streaming to get sub agent output. nests sub agent tool calls inside status containers using popovers. Agent ouptut can have any number of tool calls and AI messages imageimage


JoshuaC215:
Thanks for the contribution! I wonder, will this break or have weird behavior for other agents that use subgraphs independently outside of langgraph-supervisor?

haarisedhi102:
I tried a couple agents that i though might be impacted, but didnt see any impact. I dont think any others have subgraphs

Issue #188, opened 2025-03-12
Independent Chat Memory for Each Agent with Unique ThreadID


LounesAl:
I appreciate the effort put into this project! I found it very useful while working on my own application, and I’d love to contribute back if this improvement is relevant.

I noticed that in the current implementation, agents do not have independent memory. Each time we retrieve the conversation history, it defaults to the main agent’s memory. Additionally, when switching from one agent to another, it remains the same history.

From my perspective, it is important to separate the memory of each agent and associate chat histories with a unique threadID (uuid) especially when building agents with different purposes . This ensures that:

- Chat interactions align better with LangGraph’s principles, where each agent’s state should be self-contained. (but it depends on your app’s purpose or end goal).

Proposed Improvement :

- Avoid saving ThreadID, last message and messages in the session state directly. **Instead, use _getters_ and _setters_ in the AgentClient to retrieve the right data of the current agent.** 

E.g :  Use `st.session_state.agent_client.messages` instead of `st.session_state.messages`. This will return something like : `return self._messages[self.agent]`

- Also i used st.navigation to do multi-page agents, and in each page i set the `st.session_state.agent_client.agent = PAGE_AGENT_NAME`. This helps separate well and let you '_in the futur_' list all discussion in the agent page (Chatgpt-like)

This is something I have already implemented in my own app, and I would be happy to submit a PR in the next days if you think it would be a valuable addition to the repository.

JoshuaC215:
Hey @LounesAl, awesome! Thanks for the kind words.

In general this improvement to use separate memory for each agent in a session makes sense to me.

Because of how the streamlit runtime model operates, it's by design to keep messages in session_state while interacting with one agent, otherwise if it hits the API every time it will slow down the app execution.

I would recommend an update like the following:
- Update the `thread_id` value in streamlit app to be something like `<session_id>-<agent-name>` instead of just session_id used today
- When the selected agent changes, retrieve / reset the message history by calling the API and store it in session_state
- Otherwise everything should "just work" like today (I think)

What do you think? I would welcome a PR with that update.

Re: `st.navigation`, that sounds like fun 😄 (fun fact: I also like it because I designed `st.navigation` 😉). It might make sense to chunk this into a separate PR from the memory update for easier review and testing. But if you already have it working and want to do at once, we could, just may take me longer to review and engage with it.

Cheers! Thanks!

JoshuaC215:
[...]

PR #248, opened 2025-07-22
feature: Separate dialog history of each agent

jinhyun95:
This PR separate dialog history of multiple agents.
- As-is) Talking to multiple agents using a single user ID leads to mixed dialog history.
- To-be) Service adds agent ID to thread ID to separate dialog histories in `_handle_input`.
- Related to: https://github.com/JoshuaC215/agent-service-toolkit/issues/188

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -100,13 +100,15 @@ async def info() -> ServiceMetadata:
     )
 
 
-async def _handle_input(user_input: UserInput, agent: AgentGraph) -> tuple[dict[str, Any], UUID]:
+async def _handle_input(
+    user_input: UserInput, agent: AgentGraph, agent_id: str
+) -> tuple[dict[str, Any], UUID]:
     """
     Parse user input and handle any required interrupt resumption.
     Returns kwargs for agent invocation and the run_id.
     """
     run_id = uuid4()
-    thread_id = user_input.thread_id or str(uuid4())
+    thread_id = f"{user_input.thread_id or str(uuid4())}-{agent_id}"
     user_id = user_input.user_id or str(uuid4())
 
     configurable = {"thread_id": thread_id, "model": user_input.model, "user_id": user_id}
@@ -170,7 +172,7 @@ async def invoke(user_input: UserInput, agent_id: str = DEFAULT_AGENT) -> ChatMe
     # you'd want to include it. You could update the API to return a list of ChatMessages
     # in that case.
     agent: AgentGraph = get_agent(agent_id)
-    kwargs, run_id = await _handle_input(user_input, agent)
+    kwargs, run_id = await _handle_input(user_input, agent, agent_id)
 
     try:
         response_events: list[tuple[str, Any]] = await agent.ainvoke(**kwargs, stream_mode=["updates", "values"])  # type: ignore # fmt: skip
@@ -203,7 +205,7 @@ async def message_generator(
     This is the workhorse method for the /stream endpoint.
     """
     agent: AgentGraph = get_agent(agent_id)
-    kwargs, run_id = await _handle_input(user_input, agent)
+    kwargs, run_id = await _handle_input(user_input, agent, agent_id)
 
     try:
         # Process streamed events from the graph and yield messages over the SSE stream.


src/service/service.py
```diff
@@ -154,8 +154,10 @@ async def _handle_input(user_input: UserInput, agent: AgentGraph) -> tuple[dict[
 
 
 @router.post("/{agent_id}/invoke")
-@router.post("/invoke")
-async def invoke(user_input: UserInput, agent_id: str = DEFAULT_AGENT) -> ChatMessage:
+async def invoke(
+    user_input: UserInput,
+    agent_id: str,
+) -> ChatMessage:
```
JoshuaC215, line 160: This would mean that an agent always has to be specified - it's an API change. I'm opposed to this change.

src/client/client.py
```diff
@@ -83,6 +83,11 @@ def update_agent(self, agent: str, verify: bool = True) -> None:
                 )
         self.agent = agent
 
+    def get_agent_thread_id(self, thread_id: str) -> str:
+        if self.agent is None:
+            return thread_id
+        return f"{thread_id}-{self.agent}"
+
```
JoshuaC215, line 90: If we made this change, I believe we'd want this translation to happen in the service rather than in the client. Otherwise, an alternate client which doesn't follow this pattern could get bad / unexpected data. Right?

JoshuaC215:
I'm supportive of separating history by agent in a better way than provided today. But I'm not immediately a fan of the other changes. If you want to open an issue for removing default agent and forcing the user to always specify an agent, we could see what others in the community think and get some discussion. Probably should be separate PRs.

src/service/service.py
```diff
@@ -95,7 +95,7 @@ async def info() -> ServiceMetadata:
     return ServiceMetadata(
         agents=get_all_agent_info(),
         models=models,
-        default_agent=DEFAULT_AGENT,
+        default_agent=get_all_agent_info()[0].key,
```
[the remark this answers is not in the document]
jinhyun95: resolved in 08d79dc89f438808ceaeb57f99b4a27a5ea3e184

src/service/service.py
jinhyun95: resolved in 08d79dc89f438808ceaeb57f99b4a27a5ea3e184

src/client/client.py
jinhyun95: Thanks for the review!
resolved in 08d79dc89f438808ceaeb57f99b4a27a5ea3e184

jinhyun95:
> So I think the simplified change creates a few problems, particularly in the interaction with the streamlit app (maybe we need to add tests for those)
> 
>     1. The `/history` endpoint just looks naively for the thread_id, so I think it will always return nothing since it doesn't match the format threads are being saved in. To resolve this, we might use the same pattern used by `/invoke` and `/stream` where you can optionally insert the `{agent_id}` in the path. If it's omitted, then it returns the history of the default agent.
> 
>     2. The streamlit app session_state isn't being updated when I select a new agent, so the previous history messages will stay on the users screen even though they will no longer be visible to the agent. This is likely to confuse the user. I guess to provide persistence, we would need to query the history endpoint each time you change agents in case you're returning to an agent you talked with before.
> 
> 
> I guess we would ideally provide some tests for this functionality as well, since the tests are passing even though the user interaction behavior will be broken.

While working on  (1), I noticed that `chatbot` agent's state only retains the latest chat message due to the agent's lack of explicit state management.
Would you mind if I replace the agent with that with `MessageState`?

src/service/service.py
```diff
@@ -100,13 +100,15 @@ async def info() -> ServiceMetadata:
...
+async def _handle_input(
+    user_input: UserInput, agent: AgentGraph, agent_id: str
+) -> tuple[dict[str, Any], UUID]:
     """
     Parse user input and handle any required interrupt resumption.
     Returns kwargs for agent invocation and the run_id.
     """
     run_id = uuid4()
-    thread_id = user_input.thread_id or str(uuid4())
+    thread_id = f"{user_input.thread_id or str(uuid4())}-{agent_id}"
```
andermirik, line 111: thread_id must be uuid4 by langgraph api spec and langchain agent protocol.

JoshuaC215:
Stale, feel free to re-open if you pick it back up

PR #261, opened 2025-08-25
Pass Optional IndexConfig while initializing store

sathyarr:
* postgres may accept PostgresIndexConfig which is a sub-class of IndexConfig

### DIFF
--- a/src/memory/__init__.py
+++ b/src/memory/__init__.py
@@ -1,8 +1,10 @@
 from contextlib import AbstractAsyncContextManager
+from typing import Optional
 
 from langgraph.checkpoint.mongodb.aio import AsyncMongoDBSaver
 from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
 from langgraph.checkpoint.sqlite.aio import AsyncSqliteSaver
+from langgraph.store.base import IndexConfig
 
 from core.settings import DatabaseType, settings
 from memory.mongodb import get_mongo_saver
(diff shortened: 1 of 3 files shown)


JoshuaC215:
Hey @sathyarr — sorry for the long silence here, and thanks for sticking with it. The plumbing in the PR looks right.

On your question: you don't need the embedding *function* in `.env`. Put the model name + dims in config (just a string + int) and build the `IndexConfig` from them inside `initialize_store()`, so the call site in `service.py` stays unchanged and it's fully config-driven / off by default:

```python
# settings.py
STORE_EMBEDDING_MODEL: str | None = None   # e.g. "openai:text-embedding-3-small"
STORE_EMBEDDING_DIMS: int | None = None    # e.g. 1536

# memory/__init__.py — inside initialize_store()
index_config: IndexConfig | None = None
if settings.STORE_EMBEDDING_MODEL and settings.STORE_EMBEDDING_DIMS:
    from langchain.embeddings import init_embeddings
    index_config = {
        "embed": init_embeddings(settings.STORE_EMBEDDING_MODEL),
        "dims": settings.STORE_EMBEDDING_DIMS,
    }
```

So no placeholder needed in `service.py` — it only builds an index when those settings are set. Add the two settings to `.env.example` so they're discoverable (that's the main place to document it). If you can wire that up, take it out of draft, and add a small test that the store initializes with an index built from the settings, I'm happy to review. Thanks for your patience on this!

JoshuaC215:
Closing this out for now due to inactivity — thanks for the contribution and for working through the design questions with us. We didn't hear back after the last round of feedback.

If you'd like to pick this back up, feel free to re-open this PR (or ask for a re-open) and we'll take a look.

---

PR #262, opened 2025-08-29
Support nested sub agents

aperkins1310:
Firstly thanks, this project is super helpful and so I've made a bunch of changes that I'll submit back to upstream to try give back a bit.

The first of which is this, supporting more complex hierarchies of agents. Such as:
Screenshot 2025-08-29 at 13 37 17
This is the existing "supervisor + 2 siblings" 2-layer system:
Screenshot 2025-08-29 at 13 52 05

A few bits of note:
- The previous behaviour on the UI side relied on "finish_reason != tool_calls", which doesn't work for all models (gemini for example just uses `STOP` even for tool calls. An explicit handback message makes this work across all models. Also included tests.
- The previous behaviour on the service side was hardcoded to `supervisor`, `research_expert` and `math_expert`. Now the only requirement is supervisor graphs _contain_ `supervisor`, and sub agents contain `sub-agent` in their names. The `langgraph_supervisor_hierarchy_agent` shows how to handle the "middle" layers, which are both supervisors and sub agents. Included tests for the service behaviour.
- In order to nest statuses, we need to update streamlit to include https://github.com/streamlit/streamlit/pull/11272
- The test_service_e2e test failed locally if you have database settings that would work in the docker compose, but not out. Now we default out the settings when they're imported in `memory/__init__.py`


JoshuaC215:
This is awesome!! Thank you @aperkins1310 !!!

PR #269, opened 2025-09-17
Add a setting for log level, and graceful shutdown timeout

aperkins1310:
Log level is fairly self explanatory, graceful shutdown timeout allows ongoing requests a period of time to complete after a `SIGTERM` - which is more important for longer running requests as these generally are, [especially when running in k8s](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination)


JoshuaC215:
This is great 🙏

PR #272, opened 2025-09-27
Add generation function for improved operator ids in OpenAPI spec

nielsdebruin:
The OpenAPI spec automatically generated by FastAPI contains a unique operation ID for each route. However, the default way to generate these is `{python function name}{pathName}{method}`. This method generally does not lead to an idiomatic method name. For example, the health route will receive an operation ID of `health_check_health_get`. The problem when generating a client SDK from the OpenAPI spec is that the operation IDs are generally the same as the method name. Hence, FastAPI recommends changing them to something more idiomatic after your API is finalized [FastAPI Docs](https://fastapi.tiangolo.com/advanced/generate-clients/#custom-operation-ids-and-better-method-namesl).

This PR defines a simple generator function based on the recommended approach in the FastAPI docs that will result in idiomatic operation IDs based on the route names.

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -9,6 +9,7 @@ from uuid import UUID, uuid4
 
 from fastapi import APIRouter, Depends, FastAPI, HTTPException, status
 from fastapi.responses import StreamingResponse
+from fastapi.routing import APIRoute
 from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from langchain_core._api import LangChainBetaWarning
 from langchain_core.messages import AIMessage, AIMessageChunk, AnyMessage, HumanMessage, ToolMessage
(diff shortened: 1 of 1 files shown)


nielsdebruin:
Hi Joshua, thank you for the feedbacka and creating this template. I have added a fix to avoid the overlap when generating the OpenAPI spec.

Regarding your question about the invoke route. I am not sure....I do feel that having both might not be optimal. However, I not in favour of removing `invoke`! Instead, I would argue for removing the `{agent_id}/[invoke,stream]` and routes in favor of removing `invoke`. Why? First, given that we are creating a post request which already contains agent and model configuration it would also make sense to place the agent id in the body (The langserve api is also uses this convention I think). Moreover, when constructing a request (or set of them), all required information can be encapsulated in a single object instance, which has several benefits, e.g., making logging easier.

PR #279, opened 2025-11-19
Fix stream issue for supervisor agents

surajvariar:
Fixed the stream issue in the streamlit app for the supervisor agents.


JoshuaC215:
This is a pretty invasive change in one of the more complex parts of the code (app streaming logic). I'm hesitant to take it / not sure I'll have time to review and test thoroughly for a while.

I see langgraph-supervisor library just shipped 0.0.31 today which claims to resolve the v1 compatibility issue. Maybe try upgrading to that and reverting to the old implementation for these agents, and see if that resolves the streaming issue as a first step?

surajvariar:
have updated the version and raised a new pull request #281. I will close this one.

PR #281, opened 2025-11-21
Updated langgraph supervisor to 0.0.31

surajvariar:
Updated the langraph-supervisor to 0.0.31 which is compatible with langchain V1.


JoshuaC215:
Looked like this also fixed the stream issue 🙌

PR #284, opened 2025-11-24
Add the gemini 3 pro model to vertex.

tbenton:
This PR introduces a minor update to include the Gemini 3 Pro model in the public repository.
The previous update (see #282) added support for Google models but did not update the Vertex model. This PR completes that work by adding the missing Gemini 3 Pro configuration and ensuring model parity across providers.

**Changes include:**

- Added gemini-3-pro entry to the Vertex model list/configuration.
- Updated related enums, constants, or handlers where necessary.
- Minor cleanup to keep model definitions consistent.

This is a small, incremental update and does not introduce any breaking changes.


JoshuaC215:
Awesome, thanks @tbenton !

PR #286, opened 2025-12-09
Feat/voice input output

syasini:
Hi,

I've been working on integrating voice into Streamlit chat apps on a few projects for a while, and I'm excited about the recent addition of `accept_audio` support in v1.52.

This PR adds optional voice input/output to the Streamlit chat interface using a clean factory pattern. Users can enable speech-to-text (e.g. OpenAI Whisper or Deepgram) and text-to-speech (e.g. OpenAI TTS or ElevenLabs) using environment variables. It's completely opt-in and won't affect existing deployments (or at least that is the intention behind the design). The implementation keeps all voice logic isolated in a new `voice` module, so the core agent code stays clean.

Some quick notes about the general design:

**Architecture:**
- Factory pattern with pluggable STT/TTS providers
- All voice logic isolated in new `src/voice` module
- Streamlit dependencies confined to `VoiceManager` and core providers are UI-agnostic
- Zero impact on core agent code (technically! See the notes below on version compatibility). 

**Key features:**
- Environment based configuration (`VOICE_STT_PROVIDER`, `VOICE_TTS_PROVIDER`)
- Graceful degradation on errors (logs them but doesn't crash)
- Works with both streaming and non-streaming responses
- Easy to extend. You just need to add a new provider class

**What it looks like:**
```python
# In .env
VOICE_STT_PROVIDER=openai
VOICE_TTS_PROVIDER=openai
```
And that's pretty much it! Voice features are now enabled. We can discuss an optional toggle on the UI as well which I think would be helpful.

If the provider needs an API key, those can be passed as env variables as well. There are examples in the code. 

**About the Streamlit version bump:** Upgrading to 1.52.0 breaks some existing tests due to internal changes in `chat_input` behavior (specifically the `file_uploader_state` handling). It probably makes sense to handle the version bump in a separate PR so the core pytest suite can be updated independently.

**Unit tests:** I'll add comprehensive tests for the voice module once we align on whether this feature is a good fit for the toolkit.

Take a look at this short demo to see it in action:

https://github.com/user-attachments/assets/2ff5c842-acc8-4f22-9fa4-26e39a109e36

### DIFF
--- a/docker/Dockerfile.app
+++ b/docker/Dockerfile.app
@@ -16,6 +16,7 @@ RUN uv sync --frozen --only-group client
 
 COPY src/client/ ./client/
 COPY src/schema/ ./schema/
+COPY src/voice/ ./voice/
 COPY src/streamlit_app.py .
 
 CMD ["streamlit", "run", "streamlit_app.py"]


src/streamlit_app.py
```diff
@@ -227,8 +252,12 @@ async def amessage_iter() -> AsyncGenerator[ChatMessage, None]:
                     user_id=user_id,
                 )
                 messages.append(response)
-                st.chat_message("ai").write(response.content)
-            st.rerun()  # Clear stale containers
```
JoshuaC215: The st.rerun() is still useful, right? I'd prefer to keep it, at least for the non-voice case.

.env.example
```diff
@@ -90,3 +90,15 @@ OPENWEATHERMAP_API_KEY=
 # GitHub Personal Access Token (required for GitHub MCP server)
 # If not set, the GitHub MCP agent will have no tools
 GITHUB_PAT=
+
+# Voice Features (Optional)
```
JoshuaC215: Maybe add a comment that these need to be set on the client (app side) not the server (API) and also requires the OPENAI_API_KEY to be set on the app side.

syasini:
> Need to run pre-commit to fix formatting issues and get CI to pass. I still need to actually run and test it manually.

Sounds great. I'll do that. I totally neglected the pre-commit hook! Sorry.
I think the pytests are not going to pass until we resolve the streamlit version bump. But test it out manually and let me know what you think is the best approach to handle the update. 

Also, on a different note, we might wanna consider a toggle to activate audio generation in case we decide mid-session we don't wanna waste TTS credits anymore.

.env.example
```diff
@@ -90,3 +90,15 @@ OPENWEATHERMAP_API_KEY=
...
+# Set provider name to enable voice input/output. Leave empty to disable.
+VOICE_STT_PROVIDER=        # Speech-to-text provider (options: openai, deepgram)
+VOICE_TTS_PROVIDER=        # Text-to-speech provider (options: openai, elevenlabs)
+
+# Voice provider API keys (optional - defaults to provider's standard key above)
+# For OpenAI: uses OPENAI_API_KEY (already defined above)
+# For Deepgram: set DEEPGRAM_API_KEY if using Deepgram STT
+# For ElevenLabs: set ELEVENLABS_API_KEY if using ElevenLabs TTS
+# DEEPGRAM_API_KEY=
+# ELEVENLABS_API_KEY=
```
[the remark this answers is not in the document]
JoshuaC215: Yep, the format makes sense and I would just remove the commented lines, and specify in the comments on provider vars that the only available option for now is `openai`

.env.example
syasini: Done ✅

syasini:
@JoshuaC215 Thanks again for the review and the comments! A few updates: 
 
I added the `st.rerun()` and there was a minor issue with the audio, but now it's working fine. 
I was not adding the audio message to session_state, so after `st.rerun()` they would disappear. 
Now ONLY the last audio message is held in session state (to preserve memory), and when chat is restarted it gets deleted. 

a separate note on the AppTests: so the Streamlit app test framework has not been updated to support the new chat_input().  If you run the tests locally with STT or TSS providers (e.g. `VOICE_STT_PROVIDER=openai`) the tests will fail.  
First, I thought this was a simple Streamlit version compatibility, but I think it's a bit more nuanced than that. 
You're the AppTest expert, so I'd love to hear your thoughts on this.  

This is Claude's explanation: 

> Pytest Failures: Streamlit AppTest Incompatibility with accept_audio=True
> The Streamlit app tests were failing due to an incompatibility between Streamlit's AppTest testing framework and the accept_audio=True parameter introduced in Streamlit 1.52.0. Root Cause: When voice features are enabled via environment variables (VOICE_STT_PROVIDER=openai), the VoiceManager calls st.chat_input(placeholder, accept_audio=True). This parameter changes the widget's serialization format to support audio input alongside text. However, AppTest's internal deserialization logic hasn't been updated to handle this new format. The Error:
> 
> AttributeError: file_uploader_state
> # In streamlit/elements/widgets/chat.py:374
> uploaded_files = _pop_upload_files(ui_value.file_uploader_state)
> Why it fails:
> st.chat_input(accept_audio=True) creates a widget with ChatInputSerde(accept_files=False, accept_audio=True)
> The widget state is stored with the expectation of additional fields (file_uploader_state) to support audio uploads
> When AppTest attempts to deserialize the widget state (during on_script_will_rerun()), it fails because the test framework doesn't properly construct the ChatInputValueProto with these required fields
> The error occurs in the deserialization path, not during the initial st.chat_input() call, which is why try/except wrappers don't help
> Current Solution: Keep VOICE_STT_PROVIDER and VOICE_TTS_PROVIDER empty in .env when running tests. The VoiceManager.from_env() returns None when these aren't set, preventing the audio-capable chat input from being used. Note that pytest-env settings don't work because AppTest spawns a subprocess that doesn't inherit those environment variables and loads .env directly via load_dotenv(). Future Fix: This should be resolved when Streamlit updates AppTest to properly support accept_audio=True, or we could implement a mechanism to detect the test environment and disable audio features programmatically (though the current approach of using env vars is simpler and more explicit).
>

syasini:
@JoshuaC215 I added some unit tests with Claude to cover the voice service backend. 
The Streamlit related pieces are still not covered because of the AppTesting incompatibility. 

Let me know of you have any thoughts on this.

JoshuaC215:
Looks good! Fixed one issue with the docker setup

syasini:
Amazing! 🤩 Thanks for taking the time to review and test it, Joshua. This was really fun.

I just tried the deployed app, and it feels good to see it working there :D Just keep an eye on your OpenAI bill, those voice conversions can end up being expensive, haha. 

Hope I'll find other opportunities to contribute in the future.

PR #294, opened 2026-03-07
Switch to gpt-oss-safeguard since llamaguard was deprecated on groq cloud

JoshuaC215:


cherry-cheng:
您好，我是程永慧！您的信件已收到，我会尽快查看，谢谢！

PR #297, opened 2026-04-27
fix: prevent Anthropic "multiple non-consecutive system messages" error

### LINKED ISSUE #280: Claude fails with "Received multiple non-consecutive system messages" when using agent-service-toolkit
First of all, great work, I use this repository a lot. 

I have found that when using any of the Anthropic models (“claude-3-haiku,” “claude-haiku-4-5,” etc.), I get the following error:

ValueError: Received multiple non-consecutive system messages.

This comes from langchain_anthropic.chat_models._format_messages, which doesn't allow multiple system messages unless they are consecutive. It seems the toolkit is generating more than one system message in different positions.

A possible fix would be to merge/normalize system messages before calling the Anthropic LLM.

aizaz68:
## Summary

Fixes #280 — Claude/Anthropic models throw `ValueError: Received multiple non-consecutive system messages` when used with this toolkit.

**Root cause:** The `wrap_model` preprocessor in `research_assistant`, `rag_assistant`, and `knowledge_base_agent` prepends a fresh `SystemMessage` to `state["messages"]` on every model call. In multi-agent or supervisor setups a `SystemMessage` can already be present in the persisted conversation history, producing two system messages at non-consecutive positions — which `langchain_anthropic` rejects.

**Fix:** Filter out any pre-existing `SystemMessage` entries from the stored history before prepending the agent's own system prompt, so there is always exactly one `SystemMessage` at position 0.

## Changes

- `src/agents/research_assistant.py` — extract `_build_messages()` helper that strips existing system messages before prepending
- `src/agents/rag_assistant.py` — same pattern
- `src/agents/knowledge_base_agent.py` — same fix inside `create_system_message()`

## Test plan

- [ ] Start a multi-turn conversation using an Anthropic model (e.g. `claude-haiku-4-5`) with the `research-assistant` agent — previously raised `ValueError`, now completes successfully
- [ ] Verify same behaviour for `rag-assistant` and `knowledge-base-agent`
- [ ] Confirm OpenAI models are unaffected (filtering a non-existent type is a no-op)
- [ ] Run existing test suite: `pytest tests/`

### DIFF
--- a/src/agents/knowledge_base_agent.py
+++ b/src/agents/knowledge_base_agent.py
@@ -4,7 +4,7 @@ from typing import Any
 
 from langchain_aws import AmazonKnowledgeBasesRetriever
 from langchain_core.language_models.chat_models import BaseChatModel
-from langchain_core.messages import AIMessage, HumanMessage, SystemMessage
+from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage
 from langchain_core.runnables import RunnableConfig, RunnableLambda, RunnableSerializable
 from langchain_core.runnables.base import RunnableSequence
 from langgraph.graph import END, MessagesState, StateGraph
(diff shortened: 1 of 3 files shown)


cherry-cheng:
您好，我是程永慧！您的信件已收到，我会尽快查看，谢谢！

JoshuaC215:
Thanks for taking this on @aizaz68, and for fixing it consistently across the three agents.

Before merging I want to confirm the trigger — same thing I asked in #280, I haven't been able to reproduce the "non-consecutive system messages" error in the normal single-agent flow. Ideally a regression test that reproduces it (a state with a pre-existing SystemMessage, asserting only one ends up in the prompt) — that confirms the bug and keeps it from regressing. If a test is tricky to set up, even a manual repro would help: which agent, and what to prompt to trigger it. Once I can see the failing case I'm happy to merge. Thanks!

PR #299, opened 2026-04-27
fix: send SSE keep-alive pings during long-running agent tool calls

### LINKED ISSUE #126: Send keep-alive messages in streaming for long-running requests
In case of, for example, an agent with a tool call that is long running (e.g. more than 60 seconds), sometimes we hit network idle timeouts and it would be good to be able to send a keep alive message every N seconds, if no event comes from the agent.

aizaz68:
## Summary

Fixes #126 — when an agent runs a long tool call (e.g. >60 seconds), proxies and load balancers close the idle SSE connection before the agent responds, causing the client to see a dropped connection with no response.

## Root Cause

The `/stream` endpoint is a server-sent events (SSE) connection. During a long tool call, the agent produces no output for an extended period. Many network proxies (nginx, AWS ALB, Cloudflare) have default idle timeouts of 60s and will silently kill connections that carry no data.

## Fix

A background `asyncio` producer task feeds all agent stream events into a queue. The main generator reads from the queue with a **15-second timeout**. If nothing arrives in time, it yields an SSE comment line (`: ping\n\n`) to keep the connection alive, then resumes waiting.

```
Producer (background)          Consumer (SSE loop)
─────────────────────          ───────────────────
agent.astream() running  →  queue.get(timeout=15s)
                                   │
                         timeout? → yield ": ping"
                         event?  → process normally
                         SENTINEL → stop
```

## Why a queue and not asyncio.wait_for on the iterator directly?

`asyncio.wait_for` cancels the awaitable on timeout. When applied directly to `__anext__()`, this cancels the generator's internal coroutine and permanently breaks it — subsequent iterations raise `StopAsyncIteration` immediately. The queue approach keeps the agent generator completely undisturbed; only the queue read has a timeout.

## Cleanup on disconnect

When the user switches chat or closes the tab, the browser closes the SSE connection. FastAPI cancels the generator, which triggers the `finally` block that cancels the producer task. No orphaned tasks or memory leaks.

## Client-side impact

The SSE comment format (`: ping\n\n`) is defined in the SSE spec as a no-op. Browsers ignore it natively. The `httpx`-based `AgentClient` already only processes lines starting with `data:`, so ping lines pass through silently with no code changes needed on the client side.

## Test plan

- [ ] Trigger a long-running tool call (>60s) and confirm the connection stays open
- [ ] Confirm the client receives the final response correctly after the tool finishes  
- [ ] Switch chat mid-stream and confirm no orphaned tasks remain on the server
- [ ] Short tool calls (<15s) are unaffected — no pings sent, no behaviour change

### DIFF
--- a/src/service/service.py
+++ b/src/service/service.py
@@ -1,3 +1,4 @@
+import asyncio
 import inspect
 import json
 import logging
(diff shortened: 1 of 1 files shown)


JoshuaC215:
Hey @aizaz68, thanks for digging into this — the keep-alive approach is reasonable, and the producer-task/cleanup logic actually holds up well (I checked: a client disconnect cancels the producer, and a producer error still surfaces to the client rather than pinging forever).

**One real blocker though:** the ping breaks our own client. `AgentClient` treats any line that isn't `data: ...` as end-of-stream — `_parse_stream_line` returns `None` for `: ping`, and both `stream()`/`astream()` do `if parsed is None: break`. So when a ping fires mid-run, the client (and the Streamlit app) stops reading and the rest of the response gets dropped — the opposite of what we want. A spec-compliant EventSource would ignore the comment, but our httpx client doesn't. So this needs a client-side change too: skip lines starting with `:` and only terminate on `[DONE]`.

A couple smaller asks:
- The queue is unbounded, which loses the natural backpressure we had iterating `astream` inline — a fast agent + slow client could balloon memory. Worth a `maxsize`.
- Can you add a test or two? The wrapper's easy to test with a fake `astream` (ping-on-gap, and disconnect-cancels-producer). 

Also CI will be red on lint — `except asyncio.TimeoutError` trips ruff (UP041). If you run these and push, that'll sort it:

```sh
uv run ruff format
uv run ruff check --output-format github
uv run mypy src/
```
Let me know if the client fix gives you any trouble.

JoshuaC215:
Closing this out for now due to inactivity — thanks for digging into the keep-alive approach. We flagged a real blocker (the ping breaking `AgentClient`'s stream parsing) plus a couple of smaller asks, and didn't hear back since.

If you'd like to pick this back up with the client-side fix, feel free to re-open this PR (or ask for a re-open) and we'll take a look.

---

PR #306, opened 2026-05-25
fix: enforce AVAILABLE_MODELS validation on caller-supplied model parameter

Joshua-Medvinsky:
## Summary

The `UserInput.model` field is passed directly into the agent configurable dict without checking against `settings.AVAILABLE_MODELS`, making the operator-curated allowlist purely cosmetic.

- **Cost abuse:** callers can force expensive model variants not intended by the operator (e.g. `claude-opus-4-7` on a deployment configured for `claude-haiku-4-5`)
- **Per-request DoS:** a valid model from an unconfigured provider reaches `get_model()` with a `None` API key, returning HTTP 500 for every request
- Confirmed by `tests/service/test_service.py:74` which documents the bypass explicitly

This PR adds a validation gate in `_handle_input` that rejects any model value not present in `settings.AVAILABLE_MODELS` with a 400 response.

## Test plan

- [ ] Existing tests pass
- [ ] Sending a model value outside `AVAILABLE_MODELS` returns 400
- [ ] Sending a model value inside `AVAILABLE_MODELS` succeeds as before
- [ ] `/info` endpoint and `AVAILABLE_MODELS` enforcement are now consistent


failsafesecurity:

JoshuaC215:
Looks good. Two asks before merge: add a test covering both an out-of-list model (→ 400) and an in-list one (→ ok), and make sure `model=None` still falls through to the default.

And run a lint/type pass:

```sh
uv run ruff format
uv run ruff check --output-format github
uv run mypy src/
```

Thanks!

Joshua-Medvinsky:
Added the requested model coverage and pushed it to this branch. The tests now cover an in-allowlist model, a valid-but-disallowed model returning 400, and the `model=None` default path. I also re-ran `ruff check`, the targeted pytest cases, and `mypy src/service/service.py` successfully.

PR #307, opened 2026-05-25
fix: add TLS support for MongoDB checkpoint storage connections

Joshua-Medvinsky:
## Summary

Both MongoDB connection string paths in `memory/mongodb.py` hardcode the `mongodb://` scheme with no TLS option. `AsyncMongoDBSaver` stores full LangGraph conversation checkpoints over this unencrypted connection — every message transits as plaintext BSON.

- Affects the authenticated path (`mongodb://user:pass@host/`) and unauthenticated path (`mongodb://host/`)
- SCRAM-SHA-256 protects the auth handshake but not subsequent data
- Active only when `DATABASE_TYPE=mongo`; no `MONGO_TLS` setting exists today

This PR adds a `MONGO_TLS` boolean setting (default `True`) and appends `tls=true` to the connection URI when enabled, with a deprecation-style warning when explicitly disabled.

## Test plan

- [ ] Existing MongoDB tests pass
- [ ] New `MONGO_TLS=true` (default) produces `mongodb+srv://...?tls=true` URI
- [ ] `MONGO_TLS=false` logs a deprecation warning and omits the TLS param
- [ ] No regression on `DATABASE_TYPE=sqlite/postgres`

### DIFF
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -142,6 +142,7 @@ class Settings(BaseSettings):
     MONGO_USER: str | None = None
     MONGO_PASSWORD: SecretStr | None = None
     MONGO_AUTH_SOURCE: str | None = None
+    MONGO_TLS: bool = True  # set to False only for local development
 
     # Azure OpenAI Settings
     AZURE_OPENAI_API_KEY: SecretStr | None = None
--- a/src/core/settings.py
+++ b/src/core/settings.py
@@ -142,7 +142,7 @@ class Settings(BaseSettings):
     MONGO_USER: str | None = None
     MONGO_PASSWORD: SecretStr | None = None
     MONGO_AUTH_SOURCE: str | None = None
-    MONGO_TLS: bool = True  # set to False only for local development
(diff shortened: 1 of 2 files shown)


failsafesecurity:

JoshuaC215:
Hey @Joshua-Medvinsky — supporting TLS to Mongo is a good idea, and the connection-string handling looks correct (`&tls=true` after the existing `?authSource=`, `?tls=true` on the no-auth path).

The one blocker is the default: `MONGO_TLS=True` would break everyone running a local/docker or trusted-network Mongo without TLS the moment they upgrade. Can you flip it to opt-in (default False)? Same feature, no surprise breakage — then I'm happy to merge. Thanks!

Joshua-Medvinsky:
Good call — pushed a fix: `MONGO_TLS` now defaults to `False` (opt-in). Existing deployments won't break on upgrade; users set `MONGO_TLS=True` explicitly for production/Atlas. Thanks for the review!

PR #324, opened 2026-07-04
Implemented Chat History

surajvariar:
- Added persistent chat history support to the Streamlit app.
- Introduced a toggle to enable or disable saving chat history.
- Stored per-user history data in local JSON files under the chat history directory.
- Added a past-chats section so users can reopen previous conversations by thread.
- Updated ignore rules to prevent generated chat history files from being tracked.


JoshuaC215:
Thanks for this @surajvariar . Pulling history from the app container may be challenging since it has no persistence and gets cycled often, also if someone runs multiple app instances they won’t sync.

Would it be feasible to pull this from the existing checkpointer db on the service as we originally discussed? For example a new /history/list endpoint or similar that takes a user id and returns list of threads.

surajvariar:
closing the PR since a new one is created for chat history #358

PR #358, opened 2026-07-25
Feature/chat history

surajvariar:
Summary

1. Added new schema models and exports for chat history and thread metadata.
2. Extended the service API to return thread history and list user threads from the checkpointer.
3. Made history loading agent-aware so resumed conversations use the correct graph.
4. Updated the Streamlit app to restore previous chats from thread IDs and keep them shareable via the URL.

This is based on discussion #293.


surajvariar:
Hi @JoshuaC215, 
 Based on our discussion have implemented the chat history feature based on checkpointers. Kindly check and let me know if this approach looks good for you

src/service/service.py
```diff
@@ -425,6 +431,61 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
         raise HTTPException(status_code=500, detail="Unexpected error")
 
 
+@router.post("/{agent_id}/threads", operation_id="threads_with_agent_id")
+@router.post("/threads")
```
JoshuaC215: Should this be `get` instead of `post`?

src/schema/schema.py
```diff
@@ -173,3 +173,46 @@ class ChatHistoryInput(BaseModel):
...
+    )
+    title: str | None = Field(
+        description="Title for the thread, derived from the first human message.",
+        default=None,
+        examples=["What is the weather in Tokyo?"],
+    )
+    last_message: ChatMessage | None = Field(
+        description="Most recent AI message in the thread, used as a preview.",
+        default=None,
+    )
```
JoshuaC215: Not used, may want to drop this to shrink the response payload

src/service/service.py
```diff
@@ -425,6 +431,61 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
         raise HTTPException(status_code=500, detail="Unexpected error")
 
 
+@router.post("/{agent_id}/threads", operation_id="threads_with_agent_id")
```
JoshuaC215: This `agent_id` filter is not actually checked in the function, I think? May just want to remove this...

src/service/service.py
```diff
@@ -425,6 +431,61 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
...
+        while len(seen_threads) < input.limit:
+            page = [
+                c
+                async for c in checkpointer.alist(
+                    None, filter={"user_id": input.user_id}, before=before, limit=200
+                )
+            ]
+            if not page:
+                break
+            for tup in page:
```
JoshuaC215, line 462: Could we also add an in-line check here to confirm that the user_id does indeed match? We're assuming the checkpointer filtering works correctly, and might leak unwanted conversations if it ever breaks (maybe even a logger.warning() to complain if the check fails). And would be a very cheap check.

Also obvious but important to state: if user_id is missing from the conversation, don't return it.

tests/service/test_service.py
```diff
@@ -218,6 +218,16 @@ def test_history(test_client, mock_agent) -> None:
     assert output.messages[1].content == ANSWER
 
 
+def test_threads_without_checkpointer_returns_empty(test_client, mock_agent) -> None:
```
JoshuaC215, line 221: Need to add more robust unit tests that cover most of the relevant scenarios

tests/conftest.py
```diff
@@ -25,5 +26,14 @@ def pytest_collection_modifyitems(config, items):
...
-    with patch.dict(os.environ, {}, clear=True):
+    home_dir = (
+        os.environ.get("HOME")
+        or os.environ.get("USERPROFILE")
+        or tempfile.mkdtemp(prefix="streamlit-home-")
+    )
+    env = {
+        "HOME": home_dir,
+        "USERPROFILE": home_dir,
+    }
```
JoshuaC215: Is this change related to the PR in some way that I'm not understanding? It seems unrelated and not needed. Please explain or remove/revert this change from the PR.

src/streamlit_app.py
```diff
@@ -139,6 +139,28 @@ async def main() -> None:
                 del st.session_state.last_audio
             st.rerun()
 
+        with st.expander(":material/history: Previous Chats", expanded=False):
+            try:
+                thread_list = agent_client.get_user_threads(user_id=user_id).threads
```
JoshuaC215: This should be cached (st.cache_data) or stored in session_state, otherwise it will repeat this API call on every app update and seriously slow the app.

src/streamlit_app.py
```diff
@@ -139,6 +139,28 @@ async def main() -> None:
...
+                    try:
+                        history: ChatHistory = agent_client.get_history(
+                            thread_id=t.thread_id, agent=t.agent_id
+                        )
+                    except AgentClientError:
+                        st.error("Could not load that conversation.")
+                        st.stop()
+                    st.session_state.messages = history.messages
+                    st.session_state.thread_id = t.thread_id
+                    st.query_params["thread_id"] = t.thread_id
```
JoshuaC215, line 183: Need to also clear the audio state and update the agent based on t.agent_id

src/streamlit_app.py
```diff
@@ -139,6 +139,28 @@ async def main() -> None:
...
+            for t in thread_list:
+                label = t.title or f"Chat {t.thread_id[:8]}"
+                if st.button(label, key=f"thread_{t.thread_id}", use_container_width=True):
+                    try:
+                        history: ChatHistory = agent_client.get_history(
+                            thread_id=t.thread_id, agent=t.agent_id
+                        )
+                    except AgentClientError:
+                        st.error("Could not load that conversation.")
+                        st.stop()
```
JoshuaC215: I think we can remove the st.stop here

src/service/service.py
```diff
@@ -425,6 +431,61 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
...
+                    last_message=langchain_to_chat_message(last_ai) if last_ai else None,
+                )
+            before = RunnableConfig(
+                configurable={"checkpoint_id": page[-1].config["configurable"]["checkpoint_id"]}
+            )
+    except Exception as e:
+        logger.error(f"An exception occurred: {e}")
+        raise HTTPException(status_code=500, detail="Unexpected error")
+
+    return UserThreads(threads=list(seen_threads.values())[: input.limit])
```
JoshuaC215: can we explicitly sort by updated_at desc before returning?

src/schema/schema.py
```diff
@@ -173,3 +173,46 @@ class ChatHistoryInput(BaseModel):
...
+
+    thread_id: str = Field(
+        description="Thread ID of the conversation.",
+        examples=["847c6285-8fc9-4560-a83f-4e6285809254"],
+    )
+    agent_id: str = Field(
+        description="Agent this thread was run with.",
+        examples=["research-assistant"],
+    )
+    updated_at: str = Field(
```
JoshuaC215: This should be a datetime not a str

JoshuaC215 (changes requested):
Great improvement, some comments

tests/conftest.py
surajvariar: Sorry, this was mistake from my end. I thought i was fixing issue for one of the test case. Will revert back the changes

src/service/service.py
surajvariar: added the condition for user_id missing.

src/streamlit_app.py
surajvariar: added the state updates

tests/service/test_service.py
surajvariar: added test cases for relevant scenarios

surajvariar (commented):
Addressed the requested review comments @JoshuaC215. Kindly check and let me know if any further changes are required.

src/service/service.py
surajvariar: Changed to get

src/service/service.py
surajvariar: Removed the route with agent_id

src/streamlit_app.py
surajvariar: Added the changes to cache the api call for getting the thread

src/service/service.py
surajvariar: Added the changes

src/streamlit_app.py
surajvariar: Removed the st.stop()

src/schema/schema.py
surajvariar: Changed to datetime

src/schema/schema.py
surajvariar: Removed the last message from schema

surajvariar:
Hi @JoshuaC215 , have addressed the review comments. Kindly let me know if any further changes required.

src/service/service.py
```diff
@@ -425,6 +431,68 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
...
+async def threads(
+    input: UserThreadsInput = Depends(), agent_id: str = DEFAULT_AGENT
+) -> UserThreads:
+    """
+    List a user's conversation threads for an agent, most recently updated first.
+    """
+    agent: AgentGraph = get_agent(agent_id)
+    checkpointer = getattr(agent, "checkpointer", None)
+    if not checkpointer:
+        return UserThreads(threads=[])
```
JoshuaC215, line 445: Since [all agents share the same checkpointer](https://github.com/JoshuaC215/agent-service-toolkit/blob/main/src/service/service.py#L90-L105), doing this does not actually apply an agent_id filter - as written it will return all conversations from any agent_id.

I _think_ you could set agent_id in filter below and it would work, please test against real checkpointer backend and verify. If you want to do that, then add back the `@router.get("/{agent_id}/threads")` from before to be consistent with the other routes on the service.

src/streamlit_app.py
```diff
@@ -139,6 +150,39 @@ async def main() -> None:
...
+                        )
+                    except AgentClientError:
+                        st.error("Could not load that conversation.")
+                    st.session_state.messages = history.messages
+                    st.session_state.thread_id = t.thread_id
+                    st.query_params["thread_id"] = t.thread_id
+                    if "last_audio" in st.session_state:
+                        del st.session_state.last_audio
+                    if t.agent_id and t.agent_id != agent_client.agent:
+                        st.query_params["agent"] = t.agent_id
```
JoshuaC215: This errors in the live app, please test and fix: `StreamlitAPIException: Cannot directly set query parameter 'agent' - it is bound to a widget.`

src/streamlit_app.py
```diff
@@ -139,6 +150,39 @@ async def main() -> None:
...
+            for t in thread_list:
+                label = t.title or f"Chat {t.thread_id[:8]}"
+                if st.button(label, key=f"thread_{t.thread_id}", use_container_width=True):
+                    try:
+                        history: ChatHistory = agent_client.get_history(
+                            thread_id=t.thread_id, agent=t.agent_id
+                        )
+                    except AgentClientError:
+                        st.error("Could not load that conversation.")
+                    st.session_state.messages = history.messages
```
JoshuaC215, line 181: If the history failed to load (hit AgentClientError) then history is unbound and this will also error - please handle. Needs `st.rerun()`/`continue` after the st.error, or an `else:`

JoshuaC215 (changes requested):
Couple other issues I found in testing

src/service/service.py
surajvariar: Hi @JoshuaC215 , what is the behavior you are expecting here.

for eg: if a user starts with agent A and then switches to agent B, then the chat history of agent B should be visible not of agent A.

Is my understanding correct here?

src/streamlit_app.py
surajvariar: Addressed the comments

src/streamlit_app.py
surajvariar: Addressed the comments

src/service/service.py
surajvariar: have addressed based on my understanding

surajvariar:
@JoshuaC215 have addressed the comments provided and tested the application in local with Postgres checkpointer.

src/service/service.py
```diff
@@ -425,6 +431,76 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
         raise HTTPException(status_code=500, detail="Unexpected error")
 
 
+@router.get("/{agent_id}/threads")
+@router.get("/threads")
```
JoshuaC215: set `operation_id="threads_with_agent_id"` as [here](https://github.com/JoshuaC215/agent-service-toolkit/blob/main/src/service/service.py#L407)

src/streamlit_app.py
```diff
@@ -139,6 +150,44 @@ async def main() -> None:
...
+                        st.error("Could not load that conversation.")
+                        continue
+                    st.session_state.messages = history.messages
+                    st.session_state.thread_id = t.thread_id
+                    st.query_params["thread_id"] = t.thread_id
+                    if "last_audio" in st.session_state:
+                        del st.session_state.last_audio
+                    if t.agent_id and t.agent_id != agent_client.agent:
+                        agent_client.agent = t.agent_id
+                        fetch_user_threads_cached.clear()
```
JoshuaC215: This code won't work, but it should just be removed as it will never fire anyway - since we filter for threads by the already set agent_id, the one returned will always match

JoshuaC215 (commented):
99% there, two minor fixes. I may clean up a couple other pieces after this is merged but just fix the last two comments and its good enough to merge.

src/service/service.py
surajvariar: addressed the comment

src/streamlit_app.py
surajvariar: addressed the comment

surajvariar:
@JoshuaC215 have addressed the comments.

JoshuaC215 (approved):

JoshuaC215:
Great work! Thank you!

PR #365, opened 2026-08-12
feat(providers): add OrcaRouter as a model provider

XiaoHuo888-hue:
## What this changes

Registers OrcaRouter as a first-class model provider, mirroring the existing OpenRouter wiring so the model catalog, settings and live-check stay consistent.

- `src/schema/models.py`: adds `Provider.ORCAROUTER` and an `OrcaRouterModelName` catalog (routed model ids in `provider/model` form)
- `src/core/settings.py`: adds `ORCAROUTER_API_KEY` plus per-provider default/availability wiring
- `src/core/llm.py`: adds a `ChatOpenAI` branch pointed at `https://api.orcarouter.ai/v1`
- `scripts/check_live_models.py`: adds the provider to the live smoke sweep
- `.env.example`: documents `ORCAROUTER_API_KEY`
- `tests/core/test_llm.py` / `tests/core/test_settings.py`: unit tests for the new provider

This registers [OrcaRouter](https://www.orcarouter.ai) the same way the existing OpenRouter provider is wired, so the model picker, config reference and docs stay consistent. OrcaRouter is an OpenAI-compatible gateway: one `ORCAROUTER_API_KEY` (keys start with `sk-orca-`) unlocks 150+ models from OpenAI, Anthropic, Google, DeepSeek, Qwen, MiniMax and xAI behind a single `https://api.orcarouter.ai/v1` endpoint. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

I'm an engineer on the OrcaRouter team.

## Verification

- Unit tests: `uv run pytest tests/core/test_llm.py tests/core/test_settings.py` → 34 passed; full suite → 152 passed, 3 skipped, 1 pre-existing Windows-only failure (`tests/app/test_streamlit_app.py::test_app_simple_non_streaming` fails identically on clean `main` — Streamlit can't resolve a home directory here)
- Live API (`scripts/check_live_models.py --provider orcarouter`): all 4 catalog models PASS (`anthropic/claude-opus-5`, `deepseek/deepseek-v4-flash-0731`, `qwen/qwen3.8-max`, `grok/grok-4.3`)
- Lint/format/type: `ruff check` and `ruff format --check` clean; `uv run pyrefly check` → 0 errors

## How to use it

1. Get an API key at https://www.orcarouter.ai/console (keys start with `sk-orca-`).
2. Set `ORCAROUTER_API_KEY` in your `.env`.
3. Pick `DEFAULT_MODEL` from the `OrcaRouterModelName` catalog, e.g. `deepseek/deepseek-v4-flash-0731` (the default), `anthropic/claude-opus-5`, `qwen/qwen3.8-max` or `grok/grok-4.3`.


JoshuaC215:
Hey, thanks for putting this together, I'm not inclined to add OrcaRouter as a first-class provider though. It seems functionally very similar to OpenRouter (OpenAI-compatible gateway aggregating other providers' models), which we already support, so this would mostly be a second near-identical integration to keep in sync. I try to avoid stacking duplicate integrations like that.

It seems like OrcaRouter would be supported by just pointing COMPATIBLE_BASE_URL / COMPATIBLE_MODEL / COMPATIBLE_API_KEY from OpenRouter to https://api.orcarouter.ai/v1 and you're set, no code changes needed - is that right?

If there's more demand for a dedicated OrcaRouter entry beyond that, happy to revisit. Feel free to open an issue and let people weigh in with a 👍.

PR #366, opened 2026-08-13
Harden chat history: /threads lookup performance, titles, and coverage

JoshuaC215:
Follow-up to #358, fixing the issues deferred at merge time.

## 1. Lookup performance

`/threads` counted *threads* but paged over *checkpoints*, so the loop condition could only be satisfied by finding `limit` distinct threads. Any user with fewer threads than `limit` — the common case at the default of 20 — never satisfied it and paged to exhaustion, reading every checkpoint they own on each cache miss. `alist` is expensive per row: a separate `writes` query per checkpoint plus full msgpack deserialization of every message.

Threads are now enumerated by their **head checkpoint** (`step == -1`), which LangGraph writes exactly once per thread, then one `aget_tuple` per thread for the title and timestamp.

Measured on 6 threads × 30 turns against SQLite:

| | time | checkpoints deserialized |
|---|---|---|
| before | 436 ms | 360 |
| after | **14 ms** | **12** |

I chose this over a dedicated threads index table: the table would mean DDL and write-path changes across all three backends and new failure modes on every agent run, which is a lot of machinery for a reference template. The tradeoff is a dependency on LangGraph's step semantics — the integration tests run real graphs through a real checkpointer, so a LangGraph upgrade that broke the invariant would fail CI rather than silently drop threads.

Verified the invariant rather than assuming it: across 13 seeded threads spanning 1–30 turns and both agent styles, `step == -1` appears exactly once per thread. `Command(resume=...)` does not write a second head, and later turns continue from the last step. (`step == 1` would silently drop every single-turn thread.)

Also in this area:

- **Capped `limit`** — was unbounded, so `?limit=999999` returned 200 and forced the scan. Now `ge=1, le=100`.
- **`updated_at` is optional** — a checkpoint without a `ts` produced a 500 into a non-optional `datetime`.
- Ordering uses the tip's `checkpoint_id` (time-ordered UUIDs, always present) rather than the parsed timestamp, which avoids null and naive/aware comparison issues.

### Subgraph agents

Agents with subgraphs — both supervisor agents — write a head per subgraph call under a nested `checkpoint_ns`, inheriting the parent run's metadata. A two-turn supervisor thread produces three head rows. Those rows match the same query, so a single page covers proportionally fewer real threads and a user could silently see a fraction of their conversations.

The head lookup pages and deduplicates by `thread_id` until it has enough distinct threads or hits a hard row cap. This is resolved while paging rather than in the query because `checkpoint_ns` isn't filterable through `alist`, and the `parents` metadata key that distinguishes subgraph heads doesn't discriminate under Postgres' jsonb containment — `@> '{"parents": {}}'` matches any object, so that filter would work on Mongo/SQLite and silently no-op on Postgres.

## 2. Titles for functional-API agents

`chatbot` and `agents.py` use `@entrypoint()`, whose state lives in the `__previous__` channel, so `channel_values["messages"]` was empty and every title came back `None` — the sidebar rendered "Chat 46e793e3" for each one.

Titles now read from either channel. **I also fixed `/history` for those agents**, which was returning only the entrypoint's final AI message: a 4-turn chatbot thread returned 1 message before, 8 now. It was pre-existing, but clicking a thread in the sidebar restored a broken conversation, so leaving it would have shipped a visibly broken path. `/history` reads the raw checkpoint first and falls back to `aget_state`, keeping graph-agent behavior byte-identical at the cost of one extra indexed read on a low-frequency endpoint.

## 3. Integration tests against real backends

- New SQLite integration tests run real graph-state, functional-API, and subgraph agents through a real checkpointer.
- `/threads` coverage added to the smoke tests. Ran `postgres`, `mongo`, and `agui` — **all green, including Mongo**, whose `filter` semantics were previously unverified. The `step` filter serializes correctly through its `dumps_metadata` equality match; Postgres works via `metadata @> jsonb`.

## 4. AG-UI

`_base_config` set no `metadata`, so AG-UI runs never recorded `user_id`/`agent_id` and never appeared in `/threads`. Same metadata is now wired through, with a generated `user_id` when the client doesn't supply one, matching `/invoke`.

## 5. Test coverage

- Asserts `agent_id` actually reaches the `alist` filter — the cross-agent isolation guarantee.
- Covers the `/{agent_id}/threads` route.
- Sidebar click path, including the `AgentClientError` branch and the failed-fetch caption. This is where both of the crashes in #358 shipped, and no test exercised it.
- `test_threads_uses_requested_agent_id` rewritten onto the path route the client actually uses.
- Each new test was checked to fail without its corresponding fix.

## 6. Cleanups

- Guard asymmetry fixed: a checkpoint with no `agent_id` used to pass the inline check that exists to backstop the SQL filter.
- `aget_user_threads` added to `AgentClient` for parity with the other async methods.
- The dead `if t.agent_id and t.agent_id != agent_client.agent:` branch **isn't in `main`** — it appears to have been removed before #358 merged, so there was nothing to delete.

## 7. Docs

README documents both endpoints, the agent-scoped sidebar, the authz caveat (`user_id` is client-asserted, so any bearer-token holder can enumerate any user's threads), and the back-compat note for pre-feature threads.

## Known limits

- `MAX_THREAD_HEADS` (200 distinct threads) and `MAX_HEAD_ROWS` (1000 rows) bound the work. Threads created before the oldest head considered can fall off the list — an old thread replied to yesterday is the case to watch. This is where a dedicated index table would become the right answer.
- Tip fetches are sequential. Fine on SQLite (~0.5 ms each); bounded concurrency would help against a networked Postgres with many threads.

## Out of scope

Thread deletion/management from the sidebar.

## Unrelated issue noticed

`/interrupt-agent/invoke` returns 500 with `USE_FAKE_MODEL=true` (`'NoneType' object has no attribute 'birthdate'`) — `FakeToolModel` doesn't produce the structured output that agent needs. Pre-existing and untouched here, but it makes the keyless local dev recipe unusable for that agent. Happy to file separately.

## Verification

- `uv run pytest tests/ -q` → **175 passed, 4 skipped** (baseline 151/3; the extra skip is the new docker-marked smoke test)
- `uv run ruff check` / `ruff format --check` / `uv run pyrefly check` all clean
- `./scripts/smoke_test.sh postgres mongo agui` green, re-run after the paging change

---
_Generated by [Claude Code](https://claude.ai/code/session_015LaLmTVmBRdSArCfdeggDC)_


JoshuaC215:
## Browser verification

Drove the real Streamlit app in Chromium (Playwright) against a live service, SQLite checkpointer and the `fake` model. Seeded 5 threads for one user across `chatbot` (functional API) and `research-assistant` (graph state), mixing 1-turn and multi-turn.

**12/12 checks pass on this branch.** More usefully, I re-ran the identical script against `origin/main` to confirm the checks can actually fail — the two that this PR fixes did:

| check | `origin/main` | this branch |
|---|---|---|
| research-assistant titles render | pass | pass |
| sidebar scoped to selected agent | pass | pass |
| clicking thread restores conversation | pass | pass |
| thread_id in URL after click | pass | pass |
| no JS errors on click path | pass | pass |
| **chatbot titles render (not `Chat <uuid>`)** | **fail** | **pass** |
| **chatbot thread restores all turns** | **fail** | **pass** |

On `main` the chatbot sidebar renders `Chat b7c7256e` / `Chat c42250cc` / `Chat d9ec0bac`, and `/history` returns **1 message regardless of thread length** — a 3-turn thread came back with a single stray AI message. On this branch the same three threads render their real titles and the 3-turn thread restores all 6 messages.

No JS errors on any click path in either run.

One correction worth recording: my first attempt at the `main` comparison used the local `main` ref, which was stale at `5983303` — *before* #358 merged, so the sidebar didn't exist at all and the run was meaningless. The table above is against `origin/main` (`935318f`).

---

src/service/service.py
```diff
@@ -43,12 +49,26 @@
...
+# instead of scanning every checkpoint a user owns. Don't switch this to another step —
+# step 1 looks equivalent but single-turn threads never reach it.
+THREAD_HEAD_STEP = -1
+
+# Heads are ordered by thread creation, so over-fetch distinct threads and re-sort by the
+# tip timestamp to approximate "most recently updated". Threads created before the oldest
+# head considered can fall off the list; MAX_HEAD_ROWS bounds the work either way.
+MAX_THREAD_HEADS = 200
+MAX_HEAD_ROWS = 1000
+HEAD_PAGE_SIZE = 200
```
JoshuaC215: This is the wrong place for these constants and comments. Put it near where it’s used, not hundreds of lines away. Make the comments terse.

README.md
```diff
@@ -166,6 +167,41 @@ response.pretty_print()
 
 ```
 
+### Chat history
```
JoshuaC215: I don’t think any other APIs are documented in detail in the README? Remove, I think. Not sure if there’s a better place.

tests/service/test_threads_sqlite.py
```diff
@@ -0,0 +1,197 @@
...
+        metadata={"user_id": user_id, "agent_id": agent_id},
+    )
+    for message in messages:
+        await agent.ainvoke({"messages": [HumanMessage(content=message)]}, config=config)
+
+
+@pytest_asyncio.fixture
+async def seeded(tmp_path):
+    """Seed two agents x two users, with both single-turn and multi-turn threads."""
+    async with AsyncSqliteSaver.from_conn_string(str(tmp_path / "checkpoints.db")) as checkpointer:
```
JoshuaC215, line 74: Is it the first "mainline" test we are adding with real sqlite? is that a concern or performance issue for tests?

src/service/service.py
```diff
@@ -419,18 +439,66 @@ async def history(input: ChatHistoryInput, agent_id: str = DEFAULT_AGENT) -> Cha
...
+            state_snapshot = await agent.aget_state(config=config)
+            messages = state_snapshot.values["messages"]
         chat_messages: list[ChatMessage] = [langchain_to_chat_message(m) for m in messages]
         return ChatHistory(messages=chat_messages)
     except Exception as e:
         logger.error(f"An exception occurred: {e}")
         raise HTTPException(status_code=500, detail="Unexpected error")
 
 
+async def _list_thread_heads(checkpointer: Any, user_id: str, agent_id: str) -> list[Any]:
```
JoshuaC215: should we consider moving any of this code or other threads code out of the main service.py? is it getting to be too much? (IDK if decomposing service.py too much is also an anti-pattern from the rest of the repo structure though...)

tests/service/test_service.py
```diff
@@ -218,6 +218,95 @@ def test_history(test_client, mock_agent) -> None:
...
 
 
+class FakeCheckpointTuple:
+    def __init__(self, thread_id: str, checkpoint_id: str, checkpoint: dict, metadata: dict):
+        self.config = {"configurable": {"thread_id": thread_id, "checkpoint_id": checkpoint_id}}
+        self.checkpoint = checkpoint
+        self.metadata = metadata
+
+
+class FakeCheckpointer:
```
JoshuaC215: is it useful to maintain all this separate from the sqlite version? is it mocking too much of lang* internals?

if it is useful, should we keep it all in test_service.py or consider separating it out?

src/service/service.py
JoshuaC215: Fixed in b0c0e4f. Constants now live in `src/service/threads.py` directly above the code that reads them, and the comments are cut to the non-obvious part — kept the "don't swap in another step, single-turn threads never reach step 1" line because that's the trap I actually fell into, dropped the rest of the explanation.

---

README.md
JoshuaC215: You're right — I checked, and no other endpoint is documented in the README. Trimmed to a short pointer in b0c0e4f.

For "better place": the endpoint docstring, which the README already points at via `/redoc`. So the authz caveat and the back-compat note moved into the `threads` docstring rather than being dropped — the authz one in particular seemed worth keeping somewhere a reader hits before exposing this.

What's left in the README is 6 lines: the two endpoint names, a pointer to `/redoc`, and the one thing that isn't an API fact — that the sidebar is scoped to the selected agent.

---

tests/service/test_threads_sqlite.py
JoshuaC215: Yes, first one — `grep` for sqlite/`SqliteSaver` across `tests/` only hits this file.

Not a performance concern, measured:

```
7 passed in 0.96s
0.14s setup  test_threads_lists_single_and_multi_turn_threads   (slowest)
0.10s setup  test_threads_lists_subgraph_threads_once           (fastest)
0.03s call   test_threads_lists_single_and_multi_turn_threads   (slowest call)
```

~0.11s per test, essentially all of it in the fixture seeding 7 threads through real graphs. Full suite is ~19s, so this file is about 5% of it.

No new infrastructure either: `aiosqlite` is already a core dependency (not an extra), the DB is a `tmp_path` file so tests stay isolated and get cleaned up, and there's no Docker or network involved. It runs in CI as-is.

The thing I'd watch isn't runtime, it's the fixture — it's module-level shared setup, so it'll get slower if it keeps accreting agents. If it grows much past the current 3 agents / 7 threads I'd split it rather than let every test pay for seeding it doesn't use.

---

src/service/service.py
JoshuaC215: Agreed it was too much — done in b0c0e4f, but split along a specific line rather than moving the whole feature.

`src/service/threads.py` now holds the checkpointer mechanics (constants, head paging, summary building). The **route stays in service.py** with every other route. So the split is "endpoints in one file, the gnarly checkpointer logic beside `agui.py` and `utils.py`" rather than "a file per feature" — which I think is the anti-pattern you're wary of, since a reader looking for `/threads` still finds it where all the other endpoints are.

`service.py` 590 → 496 lines; the endpoint is now ~25 lines of routing and error handling.

The reason this particular code earns a module: it's the only place in the service reasoning about LangGraph checkpoint internals (step numbering, nested namespaces, tip vs. head), it carries four tuning constants, and it has its own test files. That's a coherent unit. I wouldn't extract `/invoke` or `/history` on the same logic — they're thin.

This also resolved your constants comment naturally: they now sit directly above the paging loop that reads them.

---

tests/service/test_service.py
JoshuaC215: Good challenge, and it was mocking too much. Fixed both parts in b0c0e4f.

**It was over-mocking.** The fake modelled the `__previous__`/`__end__` channels so it could test functional-API titles — i.e. asserting my own reimplementation of a LangGraph detail. I deleted every fake-driven test the SQLite suite already covers more faithfully:

- `filters_and_orders_from_checkpointer`, `enumerates_by_head_checkpoint` → SQLite `isolates_users_and_agents` + `lists_single_and_multi_turn_threads`
- `titles_functional_api_threads` → SQLite `titles_functional_api_agent`
- `truncates_to_limit` → SQLite `respects_limit`
- `deduplicates_threads` → SQLite `lists_subgraph_threads_once`
- `history_reads_functional_api_previous_channel` → SQLite `history_returns_full_conversation`

The fake no longer models `__previous__`/`__end__` at all.

**It's still worth keeping** for the cases a real checkpointer can't be made to produce on demand:

| test | why not SQLite |
|---|---|
| `pages_past_subgraph_heads`, `bounds_total_rows_scanned` | need 300–600 checkpoints; seeding that through real graphs to test a cap isn't worth the seconds |
| `skips_checkpoints_with_mismatched_metadata` | needs a checkpointer that *ignores* the filter — that guard exists precisely to backstop the DB |
| `tolerates_missing_timestamp` | SQLite always writes a `ts` |
| `checkpointer_error_returns_500` | needs an injected failure |
| `filters_by_agent_id` | asserts the filter dict reaching `alist`, which is the isolation guarantee itself, not just its effect |
| `rejects_out_of_range_limit`, `without_checkpointer_returns_empty` | never reach a checkpointer |

**Separated**, as you suggested: `tests/service/test_threads.py`. `test_service.py` drops 351 lines. Module docstrings on both files state which belongs where, so the next person doesn't add a filtering test to the fake.

Net: 175 → 169 tests, all of the loss redundant. I re-ran the mutation check afterwards — both cap tests still fail without the paging loop.

---

JoshuaC215:
## Where else real-sqlite coverage would pay off

Audited the endpoints for behaviour a mock can hide. Split into "belongs in this PR" and "separate PR", and one that turned out to be a non-issue.

### Added here (89cec54): AG-UI runs are listed by `/threads`

This one was a gap in this PR's own claim, so I closed it rather than deferring. `test_agui_records_thread_metadata` verified that `user_id`/`agent_id` land in the checkpoint — one level below the actual claim, which is that the resulting thread *comes back from `/threads`*. Those differ: the thread also has to produce a `step == -1` head that satisfies the filter.

New test runs through `/agui/{agent}/run` and then asserts `GET /{agent}/threads` lists it with the right title. Confirmed it fails if the `metadata=` wiring is removed from `_base_config`.

### Worth a separate PR: the `/invoke` + `/stream` interrupt-resume branch

The strongest remaining gap, and it's pre-existing rather than something this PR introduced.

`_handle_input` inspects `state.tasks[].interrupts` and, when a thread is interrupted, sends the next user message as `Command(resume=...)` instead of a new `HumanMessage`. **No test exercises that branch.** `mock_agent.aget_state` always returns `tasks=()`, and `grep` for `Command`/`resume` across the non-AG-UI service tests returns nothing. `test_invoke_interrupt` covers the *response* when an agent interrupts, not the resume on the following call.

That branch depends entirely on real checkpointer state, so it's exactly what a mock can't cover — a LangGraph change to how pending interrupts surface would pass CI today. AG-UI has `test_agui_interrupt_and_resume` against `MemorySaver`, but that goes through `forwardedProps.command` and `ag-ui-langgraph`, not `_handle_input`, so it doesn't cover this.

Shape: interrupt agent + real checkpointer, `POST /invoke` → assert interrupt returned, `POST /invoke` again on the same thread → assert it resumed rather than starting a new turn. Plus the `/stream` equivalent. Happy to open it if you want.

### Checked and not worth it

- **`/history`** — already covered against SQLite here, for both graph-state and functional-API agents.
- **`/feedback`, `/info`, `/health`** — no checkpointer involved; the existing mocks test the right thing.
- **`/stream` plumbing** — the SSE/token path isn't checkpointer-dependent. Only its shared `_handle_input` call is, which the item above covers.

### One caution on the pattern

Real-sqlite tests are cheap here (~0.11s each) because the graphs are trivial echo nodes. That stays true only as long as they don't reach for the real agents in `src/agents/` — those pull in models, tools, and network-shaped dependencies. Worth keeping the rule "integration tests build their own minimal graphs" if this spreads to other endpoints.

---

PR #377, opened 2026-09-06
feat: add Atlas Cloud as a model provider

binyangzhu000-sudo:
## What

Adds Atlas Cloud as a model provider. It's an OpenAI-compatible gateway, so this follows the **OpenRouter** provider exactly — same shape, same four touchpoints, no new dependency:

| File | Change |
|---|---|
| `src/schema/models.py` | `Provider.ATLASCLOUD`, an `AtlasCloudModelName` enum (4 ids), union member |
| `src/core/llm.py` | import, `_MODEL_TABLE` row, a `ChatOpenAI` branch pinned to `https://api.atlascloud.ai/v1/` |
| `src/core/settings.py` | `ATLASCLOUD_API_KEY`, the `api_keys` row, the `match` arm that sets `DEFAULT_MODEL` / `AVAILABLE_MODELS` |
| `.env.example` | one line |
| `tests/core/test_llm.py`, `tests/core/test_settings.py` | 3 tests, mirroring the OpenRouter ones |

The explicit key check is copied from OpenRouter's for the same reason its comment gives: without it the openai SDK falls back to `OPENAI_API_KEY` and would send that key to `api.atlascloud.ai`.

```bash
echo 'ATLASCLOUD_API_KEY=apikey-...' >> .env
uv run src/run_service.py
```

## Verification

**Checks CI runs** — all green, on top of a clean baseline:

- `uv run ruff format --check` → `81 files already formatted`
- `uv run ruff check` → `All checks passed!`
- `uv run pyrefly check` → `0 errors`
- `uv run pytest` → **193 passed, 4 skipped** (`tests/core` goes 35 → 38 with the 3 added tests)

**Against the live gateway, with `ATLASCLOUD_API_KEY` as the only provider key set** — so the settings wiring added here is what's under test:

```
DEFAULT_MODEL   = qwen/qwen3.5-35b-a3b
AVAILABLE_MODELS = ['deepseek-ai/DeepSeek-V3.1', 'moonshotai/kimi-k2.6',
                    'qwen/qwen3.5-35b-a3b', 'zai-org/glm-5']
base_url = https://api.atlascloud.ai/v1/  | streaming = True
```

Then through this repo's own graphs, not a hand-rolled client:

```
[chatbot]            messages=1  tool_calls=[]            last='hello from atlas'
[research-assistant] messages=4  tool_calls=['Calculator'] last='The result of 4321 multiplied by 1234 is **5,332,114**.'
```

`bind_tools` and token streaming were checked directly too: a bound tool came back as `multiply {'a': 23, 'b': 19}`, and `.stream()` yielded 1366 chunks — the two things that matter for `/stream` with `stream_tokens=True`.

## Model choice

The four ids are the ones I actually exercised with tool calling; `qwen/qwen3.5-35b-a3b` is the default because it's the cheapest of them that still drives the agent loop reliably. Any of the gateway's other models can be reached today through the existing `OPENAI_COMPATIBLE` slot — this entry just makes the common ones selectable by name, the way `OpenRouterModelName` does.

## Disclosure

I work on Atlas Cloud. This is a plain OpenAI-compatible provider used under the user's own key: nothing is reachable unless someone sets `ATLASCLOUD_API_KEY`, and no existing provider's behavior changes. Happy to trim the enum, change the default, or drop this to docs-only if you'd rather not carry another named gateway.


JoshuaC215:
Hey @binyangzhu000-sudo, thanks for the thorough write-up, and for flagging the Atlas Cloud affiliation up front — appreciated.

I went through this same reasoning recently with OrcaRouter (#365): Atlas Cloud is an OpenAI-compatible gateway, so it's already reachable today via `COMPATIBLE_BASE_URL=https://api.atlascloud.ai/v1/`, `COMPATIBLE_MODEL`, and `COMPATIBLE_API_KEY` — no code change needed. I'd rather not carry a second near-duplicate of the OpenRouter integration just to make specific model IDs selectable by short name.

If there's real demand for a dedicated entry beyond what COMPATIBLE_* covers, happy to revisit — feel free to open an issue and let people weigh in with a 👍.

PR #380, opened 2026-09-11
fix: keep SSE agent streams alive during quiet periods

### LINKED ISSUE #126: Send keep-alive messages in streaming for long-running requests

123342wra:
## Summary
- keep SSE agent streams alive during quiet periods
- bound the event queue and propagate producer failures
- clean up producers when clients disconnect
- accept SSE heartbeat comments in sync and async clients
- add regression tests

## Testing
- pytest tests/service tests/client
- full non-UI test suite

Fixes #126

### DIFF
--- a/src/client/client.py
+++ b/src/client/client.py
@@ -1,6 +1,7 @@
 import json
 import os
 from collections.abc import AsyncGenerator, Generator
+from enum import Enum
 from typing import Any
 
 import httpx
@@ -20,6 +21,11 @@ from schema import (
 
 class AgentClientError(Exception):
     pass
+
+
+class _StreamControl(Enum):
+    IGNORE = "ignore"
+    DONE = "done"
 
 
 class AgentClient:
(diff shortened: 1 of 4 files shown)


JoshuaC215:
Thanks for picking this up! Having the client handle `: ping` is a good catch too, since the current client ends the stream on any non-`data:` line.

A couple of things before this can go in:

1. **Every stream now waits one full keepalive interval before `[DONE]`.** The loop only checks `producer_task.done()` before it waits on `queue.get()`. With a real graph, the producer is still finishing up (checkpoint writes etc.) after its last event, so the consumer waits the full 15s timeout before it notices. Agent errors are delayed the same way. You can see it in the existing real-graph tests: they still pass, but `tests/service/test_service_real_graphs.py` goes from ~0.3s on main to ~75s on this branch, with each `/stream` call taking exactly 15s longer. The fix is to have the producer put an end-of-stream sentinel (or the exception it caught) on the queue in a `finally`, and have the consumer stop when it sees that, like #299 did.

2. **Keep alive based on what the client sees, not on agent events.** At the moment the ping timer resets on any `agent.astream` event, but `message_generator` drops a lot of those, e.g. every token when `stream_tokens=false`. A long LLM call with token streaming off would send no pings and could still hit the idle timeout. I think it's cleaner to wrap the SSE output instead, something like `StreamingResponse(_with_keepalive(message_generator(...)))`. That leaves `message_generator` unchanged and removes the ping special case.

3. **Add a regression test** in `test_service_real_graphs.py`: stream a real graph with the interval patched down, and assert that `[DONE]` arrives well before the interval. A second test with a slow node should check that a ping comes through.

The client-side changes look good. Should be good to merge after those!

PR #381, opened 2026-09-11
feat: add local knowledge base RAG agent

123342wra:
## Summary
- add local ingestion for TXT, Markdown, PDF, and DOCX files
- persist chunks in per-knowledge-base Chroma collections
- add local-rag-agent with source/page citations
- add sync/async client methods, Streamlit upload UI, and regression tests

## Testing
- pytest tests --ignore=tests/app --ignore=tests/integration -q
- ruff check .
- ruff format --check .
- pyrefly check
- python -m compileall -q src tests

`USE_FAKE_MODEL=true` supports offline tests; real embeddings use OpenAI
`text-embedding-3-small`.


JoshuaC215:
Hey @123342wra, thanks! This overlaps pretty heavily with the existing rag_assistant + Chroma setup (docs/RAG_Assistant.md), and adds a fairly large new surface on top of it in one PR: a new knowledge module, two new endpoints, a new dependency, and Streamlit upload UI.

I'd rather not take this on as-is. If you want to pursue it, can you open an issue first laying out why the existing RAG assistant isn't enough so we can agree on scope before more code gets written? Happy to give feedback there.

PR #382, opened 2026-09-16
Deepagents supervisor migration

haarisedhi102:
replaces langgraph-supervisor (used by langgraph-supervisor-agent and langgraph-supervisor-hierarchy-agent) with deepagents, which expresses the same supervisor/subagent pattern through a single task tool. Both agents behave the same; the hierarchy's inner research team is now a nested deep agent instead of a nested create_supervisor.


JoshuaC215:
Hey @haarisedhi102 can you provide some more context on this proposed change? I think migrate to deepagents makes sense, are the rest of the changes also needed? Thanks for the contribution!

JoshuaC215:
Also seems to mix an interrupt fix and new model addition in the same PR, could we separate out the distinct pieces for easier review unless they need to be tied together? Thanks!

haarisedhi102:
Hi @JoshuaC215 removed the interupt changes, everything else is related to migrating the transfer_* tools from the legacy supervisor lib to the task tool that deepagents uses

